PR #44, opened 2015-01-14
support for IE7/8

BenK22:
appendChild and insertRule calls throw errors in IE7/8.


troolee:
Can you please rename commits to have more meaning instead of just "Update gridstack.js"?

BenK22:
Yeah, I'm still in the process of figuring git out.

BenK22:
I deleted then created a new fork with the changes I've made (AMD and IE7/8 support) as renaming commits doesn't look possible.

PR #51, opened 2015-01-28
Get grid coordinates from a pixel position

juchi:
Hi,
I noticed that there is no direct way to insert a widget from a mouse click or drag&drop of an outside DOM element, as it's needed to resolve the coordinates of the future widget in grid coordinates (row, column).

This PR add a new method get_cell_from_pixel() which will return the row and column corresponding to a pixel coordinates. The input parameter can be directly taken from a drop event from jQueryUI or a mouse event, as in both case they are available as document coordinates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,16 @@ grid.cell_height(grid.cell_width() * 1.2);
 
 Gets current cell width.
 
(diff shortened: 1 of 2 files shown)


troolee:
Thank you!

Issue #60, opened 2015-02-04
Changing number of columns dynamically


PhilipAnthonyMurray:
Is it possible to change the number of column using data-gs-width and have the grid layout update the contained widgets?

troolee:
It's impossible now. I will implement this later

radiolips:
@PhilipAnthonyMurray This functionality has been added and is available in v0.2.5. Only took a year and two weeks! Apologies for the delay.
#227

PR #72, opened 2015-02-19
Change the number of columns dynamically

juchi:
Hi,

Add the ability to use the "width" option, which was not usable if i'm right since we were forced to manually update the css. I have added a new function _update_horizontal_styles() in order to manage this, and therefore removed the related css rules that become useless.
Please note that for dynamic styles I have used the attributes proposed in my previous PR (cssRules and deleteRule).

This partially solves issue #60 though it's only used at init step, the number of columns cannot be updated afterwards but it's a first step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,38 +426,6 @@ and HTML:
 
 See [example](http://troolee.github.io/gridstack.js/demo/knockout.html).
 
-
-## Change grid width
-
-To change grid width (columns count), to addition to `width` option, CSS rules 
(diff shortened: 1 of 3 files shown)


troolee:
It's not a good idea to define width rules inside JS code. As soon as they can be either responsive or static. 

To implement different column size it's better to use class names and define different column width setup for different classes.

I will provide a sample soon.

dushyant:
Is this resolved? I have similar requirement to allow user to switch between two and three columns. Is this possible to change grid.options.width?

PR #87, opened 2015-03-04
switch widget 

lorenzopub:
hi troolee,

first of all, thanks for this awesome library! i'm using it in a project to create dashboard.

currently, when we drag a widget in the grid, others widgets are moving accordingly, that fine.

but in some case i would like to have a different behavior, like switching 2 widgets, each one taking the size and positionning of the other one.

let me know if it sound relevant to you

cheers
lorenzo


troolee:
Hi Lorenzo,

Do you need an option for each widget of for whole grid? Please describe your use case.

Could you please open a ticket instead of Pull Request.

PR #140, opened 2015-05-03
added resize_grid() method, n allowed multiple classes as the 'item_class'.

rosspi:
1. Added a resize_grid() method where the nodes will retain their old positions when the grid is resized.
2. And split up the use of the item_class property between using it as a jquery selector, and using it to build some html - via a '_item_class_selector'.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -706,13 +706,15 @@
                     node.height = preferredH;
                 }
                 node._dirty = true;
+
                 node = this.grid._prepare_node(node, false);
+                this.grid._fix_collisions(node);
+                
             }, this);
 
-            this.grid._fix_collisions(node);
             this.grid._pack_nodes();
             this.grid._notify(); 
-            
+
             this._update_container_height();
         }
     }; 
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -707,11 +707,12 @@
                 }
                 node._dirty = true;
                 node = this.grid._prepare_node(node, false);
-                this.grid._fix_collisions(node);
-
-                this.grid._pack_nodes();
-                this.grid._notify(); 
             }, this);
+
+            this.grid._fix_collisions(node);
+            this.grid._pack_nodes();
+            this.grid._notify(); 
+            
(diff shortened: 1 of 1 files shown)


src/gridstack.js
```diff
@@ -379,7 +383,8 @@
         this.opts = _.defaults(opts || {}, {
             width: parseInt(this.container.attr('data-gs-width')) || 12,
             height: parseInt(this.container.attr('data-gs-height')) || 0,
-            item_class: 'grid-stack-item',
+            item_class: opts.item_class,
```
abologna-r7: There should be a default class, here what you're doing is that you're enforcing to pass item_class, and if it's not passed it will throw an error. Same for line below \/

abologna-r7:
I much needed feature :D

src/gridstack.js
papalaza: That's not true because two lines above this change is:
opts.item_class = opts.item_class || 'grid-stack-item';

abologna-r7: Got it! I didn't see the full spectrum of the file ;)

papalaza:
Hi troolee, 
You would know best.
I'll make that change.

papalaza:
@troolee  Actually, the fix collisions call requires a node as a param, so I must do this within the each loop, right?

cfresh:
I have been using the code in this PR to re-layout a grid with fixed-width columns on window resize, and found that fixing collisions in the same loop as you use to update the position and size causes the panels to swap positions and change rows seemingly at random. Adding a second loop to fix collisions after the main loop remedied this situation.

JohnBlazek:
@cfresh can you share your code?  This fix seems to address part of what i need to do, however, it doesn't maintain order when expanding the browser.

radiolips:
We've merged in a not-so-dissimilar change set. It doesn't include the original value store, but in an upcoming update it will. Thanks for coming up with this work originally - your work is very much appreciated. Please continue to add PRs where you see fit, we'd love to see more code contributed in the future!

PR #147, opened 2015-05-19
_is_one_column_mode unreliable within five pixels

radiolips:
$(window).width() is incorrect by about 5 pixels, causing _is_one_column_mode to be incorrect. To test this, go to any gridstack and make the page width 769px. Single column mode via CSS will not kick in, yet elements will not be draggable.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -550,7 +550,7 @@
     };
 
     GridStack.prototype._is_one_column_mode = function () {
-        return $(window).width() <= this.opts.min_width;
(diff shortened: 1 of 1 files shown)


radiolips:
This can be seen even on http://troolee.github.io/gridstack.js/ - make the page just over 768px (769, 770) and try to drag-and-drop the elements. I am testing this on Mac - Chrome, but I would guess it happens on most (all) browsers.

radiolips:
Upon further testing, this appears to only happen with certain resolutions - including Thunderbolt display and another high-dpi 27" monitor.

PR #175, opened 2015-06-30
Set cell_height from cell_width if not provide.

namerci:
Prevent to crash when we set cell_height to null. Also is helpful if we need square blocks.
.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -465,7 +465,6 @@
 
         var on_resize_handler = function() {
             if(is_no_cell_height) {
-                console.log('no cel height');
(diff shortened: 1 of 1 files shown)


pySilver:
@namerci there is console.log left

namerci:
Sorry

derekm:
I tried to implement this today, wanting square cells that scale proportionally as the window is resized. I'd like to resurrect consideration of this feature.

troolee:
I added `'auto'` value for this case. If it's 0 — cell height is taken from CSS

Issue #167, opened 2015-06-15
onchange(items) event not fired


FredFaust:
The event is not fired when calling remove_widget with floating disabled. Let's say I have multiple widgets all in one column and I call remove_widget for the top widget. Since floating is disabled, all the widgets are pushed to the top, their attributes are changed (data-gs-x and data-gs-y) but the event about widgets changing position isn't called.

I think it shouldn't be hard/long to fix but I don't understand the code well enough to do the change myself.

Thank you

mgwidmann:
:+1: I'm not seeing the `change` event fired on add or remove. Either it should do a `self.container.trigger('change', [self.grid.get_dirty_nodes()]);` or trigger an add/remove specific event.

piwi91:
:+1: I miss this too

piwi91:
@troolee  is there a workaround for this issue?

PR #190, opened 2015-07-22
Trigger change event for add/remove

cvillemure:
Trigger change event for add_widget / remove_widget methods and refactor
the trigger into a method for uniformity

Fix issue #167


piwi91:
Thank you for this one! :)

Issue #142, opened 2015-05-08
resizestop event issue


lopezcjf:
Hi,

After resizing an element I need to catch the new width / height. For that purpose I'm using the resizestop event but I couldn't find the new values in here (only the old ones)

It seems to be that the values are updated somewhere after this event , and I only can get the new values on the next resize event (which of course is not suitable)

Thanks

troolee:
Why don't you use `onchange` event? An `onchange` handler receives all affected items.

m5n:
This seems to have changed between 0.2.2 and the current 0.2.3-dev. Formerly, the new values were accessible, now the old values are returned.

Using onchange is cumbersome as it contains details about all items that changed because of a resize, rather than just the one of interest.

lopezcjf:
I guess I agree with m5c in this. It is good to support each event type with it's relevant data, specially for complex scenarios.

m5n:
I played with 0.2.3-dev a bit more, and I was wrong, the new values do get sent. (I have some other code that has a different outcome from 0.2.2, still investigating that.)

The "ui" object contains the originalSize and size properties that give you what you need:

    $('.grid-stack').on('resizestop', function (event, ui) { ... }

lopezcjf:
I swear I examined both objects and found the old values in both. But mistakes can happen! I will re-check that tonight ;)

troolee:
It could probably be because of animation. Please check it with and without.

mkez00:
I can confirm what lopezcjf is saying.

```javascript
$('.grid-stack').on('resizestop', function (event, ui) {
console.debug(ui.element[0].attributes.getNamedItem("data-gs-width"));
console.debug(ui.element[0].attributes.getNamedItem("data-gs-width").value);
});
```

first console output is: data-gs-width="1"
second console output is: 12

First console output is the actual new width of widget and second is the old value.  I am not using animation.

derekm:
See also issue #133, where there is a fix for this posted.

radiolips:
Closing, as @derekm's code has been merged. @lopezcjf please reopen if you're having a problem still.

PR #200, opened 2015-08-06
fix ordering of draggable and resizable event callbacks

### LINKED ISSUE #159: Fire onresizestop after complete resizing
When resizing a widget, the dimensions snap to the grid after the handle is released. This is great. :)

screenshot from 2015-06-09 11 56 49

However, it seems that onresizestop() is triggered before that process is finished so if you try to get the widgets dimension, you get them 'wrong'.
Is there a way to get an event that is triggered after the full resizing process?

derekm:
fixes #159 by making grid item callbacks actual event callbacks instead of options callbacks

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -677,6 +677,12 @@
             self._trigger_change_event();
 
             self.grid.end_update();
+
+            var nested_grid = o.find('.grid-stack');
+            if (nested_grid.length && event.type == 'resizestop') {
+                nested_grid.data('gridstack').on_resize_handler();
+                nested_grid.find('.grid-stack-item').trigger('resizestop');
+            }
         };
 
         el
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -660,6 +660,10 @@
 
             el.resizable('option', 'minWidth', cell_width * (node.min_width || 1));
             el.resizable('option', 'minHeight', self.opts.cell_height * (node.min_height || 1));
+
+            if (event.type == 'resizestart') {
+                o.find('.grid-stack-item').trigger('resizestart');
+            }
         };
 
         var on_end_moving = function(event, ui) {
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -686,16 +686,16 @@
         };
 
         el
-        .draggable(_.extend(this.opts.draggable, {
-            containment: this.opts.is_nested ? this.container.parent() : null
-        }))
-        .on("dragstart", on_start_moving)
-        .on("dragstop", on_end_moving)
-        .on("drag", drag_or_resize)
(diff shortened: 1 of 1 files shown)


derekm:
these fixes impact nest grids, I'm working on a fix for it

derekm:
this would also fix #142

troolee:
Thank you!

I'm on vacation now. Will apply the PR next week.

derekm:
I had put the stop-propagate calls in the wrong place. I'll have more commits today to sort things out.

This also opens a small can of worms related to triggering events between grids (triggering resize events in inner grids when its container is resized in the outer grid, etc.).

derekm:
The above commits put stop-propagation into the correct order. I'm also broadcasting "resizestop" events to nested grids.

Question: Should I broadcast all resize events to nested grids? I, personally, only need resizestop, but I can see where people might need any of the resize events to percolate thru nested grids.

I trigger the resizestop handlers individually so the grid's resizestop events don't fire more than once.

Before broadcasting resizestop, I call the window resize handler `on_resize_handler()`, because I'm using square cells where `cell_height()` is set to `cell_width()`. For others, calling the window resize handler might be a no-op or a wasted effort.

derekm:
Scratch what I said about triggering handlers individually. That idea didn't work out as expected. Grid item's in nested grids need to propagate their events normally, in case grid-level handlers will operate on the grid item via `event.target`.

bmatthews99:
Thanks for this derekm, this fixed my issue. I look forward to seeing it added to the repo.

derekm:
So I've been using this branch with much success in a dashboard product with nested grids.

Yesterday, though, I realized there's an issue with broadcasting "resizestop" events to nested grids. Grid items have a _dirty status that defaults to "true", which is the crux of the problem, and clean_nodes() is used to set all items to "false" before beginning updates.

An outer item that has a nested grid and is being resized may need to broadcast the resizestop event to nested grid items so they can repaint to their new sizes. But since only resizestop is broadcast downwards, the dirty nodes are not "cleaned", and since resizestop fires the "change" event, the change event fires as many times as there are items in the inner grid.

I'm looking for a fix to this that doesn't involve wholly refactoring how nodes are marked dirty, although it is seeming dangerous to default the dirty flag to "true."

derekm:
Broadcasting "resizestart" to inner grids appears to be enough to squelch the proliferation of unwanted "change" events without any undue effects. I'll update this pull request with these improvements soon.

derekm:
The branch has been working great for us for a while now... I think it is safe to consider this for inclusion in gridstack.js proper.

joemorcullo8:
Hi do you have any sample for this? I got lost on what exactly the solution for this. thanks :)

PR #203, opened 2015-08-08
performance and other fixes

### LINKED ISSUE #177: Massive slowdown for larger grids: _fix_collisions and move_node executed too many times
I'm working with grids that could potentially have 50+ widgets in them. When I tested with large grids, my page got really slow and started getting really heavy on my CPU. The problem is that the _fix_collisions and move_node functions get executed a LOT of times. It seems to increase exponentially as the number of items in the grid increases (I'm testing a grid with 125 widgets, and _fix_collisions is being called 700,000 times just to load the page).
I wanted to make sure it wasn't an issue with something I was doing, so I counted the number of times it is executed in the gridstack demo here: http://troolee.github.io/gridstack.js/, and it was still executing _fix_collisions a lot more than I would expect. After moving widget "1" from the top left to the bottom right of the grid, _fix_collisions had been executed 3125 times.
Do you have any thoughts on how the algorithm could be changed so that _fix_collisions is executed a whole lot less, and doesn't increase so dramatically as the number of widgets increases? Or is there no  way around executing it so many times?

h-nazzal:
the will fix #177.
also fix drag handle option being hard coded.
and i finally added some sort of limit to when the collision detection is fired when dragging a widget. now collision detection will only work when you move/drag the widget an amount equal to one unit.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -68,7 +68,6 @@
         this.width = width;
         this['float'] = float_mode || false;
         this.height = height || 0;
-
         this.nodes = items || [];
         this.onchange = onchange || function() {};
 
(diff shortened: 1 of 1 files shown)


mgwidmann:
Is there any way the fix for #177 can be separated from the rest of these changes?

h-nazzal:
@mgwidmann sorry but i dont have any time to seperate the changes into seperate commits (and i have actually forgoten the changes that have been made a month ago) i didnt get any feed back on this. so im rolling my own plugin now. as the devlopment pace here seems to be very slow.

any how to fix the performance all i had to do is correct one line which is 

````
-            nn = {x: 0, y: node.y, width: this.width, height: node.height};
+            nn = {x: node.x, y: node.y, width: node.width, height: node.height};
````

so if you want the perfornace fix only you only have to replace `_fix_collisions` with the one from my fork :)

troolee:
Unfortunately it doesn't work. Performance is much better, but if you create a lot of items and then drag a widget through the rest they change their positions even if you return the widget to its original position.

mgwidmann:
Why was this merged if it doesn't work? Am I reading your last comment wrong or is master broken? :confused:

troolee:
It was accidentally. I had to rollback to revision before the merge.


Regards,
Pavel Reznikov

pashka.reznikov@gmail.com

mgwidmann:
:+1:

orgilor:
regradless the whole commit.
chaging only this:

-            nn = {x: 0, y: node.y, width: this.width, height: node.height};
+            nn = {x: node.x, y: node.y, width: node.width, height: node.height};

resolve the performance issues without any other problems. checked this with 210 widgets before and after.

maerten:
orgilor's fix above has solved the freezing-on-drag issue for me

PR #214, opened 2015-08-20
Placeholder takes up space when clearing the grid

mgwidmann:
In our app we are clearing the grid and then reprinting the DOM and telling gridstack to reinitialize. When that happens, the placeholder div is still around and gets moved up to the top 0, 0 position. After reinitializing, elements get pushed down because of the placeholder (even though it is display: none).

This PR simply makes sure not to look for placeholders when calling `_prepare_element`.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -446,7 +446,7 @@
         if (this.opts.auto) {
             var elements = [];
             var _this = this;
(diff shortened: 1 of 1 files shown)


PhilipAnthonyMurray:
+1

Issue #174, opened 2015-06-28
min_width needs better docs


pySilver:
Hi!,
Docs says:
- `min_width` - minimal width. If window width is less, grid will be shown in one-column mode (default: `768`)

However, thats not true. Without overriding CSS media query, lesser values won't be taken into account (yeah, grid will be locked, but it still goes one column mode earlier than expeced)
```css
# responsible line
@media (max-width: 768px) {}
```

I'm not really sure, but something like might help:

```javascript
document.querySelector('style').textContent +=
    "@media (max-width: 768px) { .grid-stack-item { /*undo styles?*/ }}"
```

josephndenton:
Saw this issue as well. Was able to fix it to generate the media query based on the value of min_width with:

```css
/* Comment this portion out of the CSS */
/*
@media (max-width: 768px) {
  .grid-stack-item {
    position: relative !important;
    width: auto !important;
    left: 0 !important;
    top: auto !important;
    margin-bottom: 20px;
  }
  .grid-stack-item .ui-resizable-handle {
    display: none;
  }

  .grid-stack {
    height: auto !important;
  }
}
*/
```

```javascript 
GridStack.prototype._update_styles = function (max_height) {
...
    // Add this line below to the bottom of GridStack.prototype._update_styles to update the media query to the value of min_width
    Utils.insert_css_rule(this._styles, '@media (max-width: ' + this.opts.min_width + 'px)', '.grid-stack-item { position: relative !important; width: auto !important; left: 0 !important; top: auto !important; margin-bottom: $vertical_padding; } .grid-stack-item .ui-resizable-handle { display: none; } .grid-stack { height: auto !important; }', 0);

};
```

henryw14:
This is a bug. I was just about to open the same PR. The media query that is putting it into one column mode:
```
@media (max-width: 768px) {
  .grid-stack-item {
    position: relative !important;
    width: auto !important;
    left: 0 !important;
    top: auto !important;
    margin-bottom: 20px;
  }
  .grid-stack-item .ui-resizable-handle {
    display: none;
  }

  .grid-stack {
    height: auto !important;
  }
}
```
should be driven by the passed in min_width param or the docs need to change. Also, this doesn't really make much sense to me:
```
GridStack.prototype._is_one_column_mode = function () {
        return (window.innerWidth || document.documentElement.clientWidth || document.body.clientWidth) <= this.opts.min_width;
};
```

especially given that the docs say:
[...]

PR #223, opened 2015-09-03
Dynamic min_width instead of hard coded

josephndenton:
Implemented the fix I mentioned in #174.

Removed the hard coded media query and makes it dynamic based on the value of 'min_width' that can be set when gridstack in instantiated.


troolee:
I don't think it's a good way to do. Better to update CSS file with corresponding value. I've updated README

PR #272, opened 2015-11-03
Remove widget: When data doesnt exists try to find the el in nodes of the engine

marcodejongh:
Hi There,

I'm working on a meteor packaged version of this plugin now everything mostly works fine. 
Except I can't get ```remove_widget``` to work because Meteor is a bit too aggressive with cleaning up the data of the elements.
I've created https://github.com/meteor/meteor/issues/5573 for that but it probably will take forever for the Meteor team to look into it.

The workaround I came up with isn't really damaging or wrong and actually makes the code more robust.
If for some reason the widget you try to remove doesn't have the ```_gridstack_node``` data it tries to look up the node in the nodes array of the engine. You might even consider adding the lookup to all methods. Although the data not existing will probably only ever occur in the ```remove_widget``` function

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -86,6 +86,10 @@
         this.float = true;
     };
 
+    GridStackEngine.prototype.getNodeDataByDOMEl = function(el) {
+        return _.find(this.nodes, function (n) { return el.get(0) === n.el.get(0); });
+    };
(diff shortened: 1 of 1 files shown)


troolee:
I've added this code to master

PR #283, opened 2015-11-18
Gruntfile to generate dist files

jlowcs:
Does the following:
- processes SCSS files
- minifies CSS files
- copies JS file
- uglifies JS file

Just run ``npm install``, then ``grunt``.

I also removed the CSS files in ``src``, since they are dist files.

Also added a ``.gitignore``.

I checked that it was still working using the demos.


marcodejongh:
:+1:

PR #326, opened 2016-02-15
new option: drag_constraint

jlowcs:
This is to ignore the dragging outside of a container. It is not the same as the constraint of jquery draggable, as that one forbids dragging outside of the container. Here, one can drag outside of the container, but the widget position will not be impacted by that dragging, ie the placeholder will go back to its initial position.

This is useful when one wants to add behavior when dragging on another part of the app. For instance, dragging on a menu, in order to move the widget to another dashboard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ $(function () {
 - `static_grid` - makes grid static (default `false`). If true widgets are not movable/resizable. You don't even need jQueryUI draggable/resizable.  A CSS class `grid-stack-static` is also added to the container.
 - `vertical_margin` - vertical gap size (default: `20`)
 - `width` - amount of columns (default: `12`)
+- `drag_constraint` - container to which the drag will be constrained. If dragging outside of it, the move placeholder will be set back to its initial position. Useful if one wants to add behavior when dragging onto other elements of the app.
 
 ## Grid attributes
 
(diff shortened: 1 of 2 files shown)


radiolips:
@jlowcs Have you tried passing the `containment` option to the draggable option? I haven't tested it yet, but it should work. draggable options are passed directly to jquery-ui's draggable, which supports this. http://api.jqueryui.com/draggable/#option-containment

jlowcs:
What draggable's containment option does is not what I want to achieve.

http://www.pureexample.com/jquery-ui/draggable-containment.html

The drag is contained in the sense that you *can't* drag outside. I want to be able to drag outside, but I want the grid to ignore what happens outside of the container (ie, not move the widget placeholder).

Maybe the name I chose for the option is not ideal.

radiolips:
Ah, of course, my mistake. Thanks for the explanation - the option name had me confused, you're right.

In the very near future (within this week, hopefully), @troolee will have add this functionality plus more, to enable widgets to be dragged out of the grid, onto another grid, or onto another object (think: trash). I'm going to decline this PR, but only because the code will be added. Thanks for the suggestion and for providing the code - I'll ensure that this functionality is added and that you're credited.

jlowcs:
:+1:

PR #328, opened 2016-02-16
Fix/skip pack nodes on node removal

dstarchevskyy:


troolee:
PR contains too many changes. Goals of some of them are unclear. Some changes break API interface.

Issue #216, opened 2015-08-21
Destroy shouldn't delete DOM


henryw14:
Hey, great plugin. My major issue with it is that the destroy method actually deletes the DOM. I was hoping it would merely uninit the dashboard. I am using Ember.js and here's my use case. When adding a widget, I simply added it to an array which causes a rerender which adds another widget into the dashboard (thanks to two-way binding). The problem is that this new widget isn't considered a part of the dashboard even though it has all of the correct data attributes. So I thought I'll do what I do with other plugins. Call destroy, the init again. Only, I lost all of my DOM :( 

See, your `add_widget` method is not conducive to Ember because it requires that I pass html to the helper method itself, which isn't at all what I want to do from an Ember perspective. It'd be nice to either not destroy the DOM or provide a way to add a widget to the dashboard that already exists in the DOM to the dashboard container. As in, don't programmatically create a new widget for me. Let me specify that this widget doesn't currently exist in the dashboard but it should, so add it.

edborden:
I use Gridstack with Ember and use `add_widget` in setup like this:
`didInsertElement: -> scheduleOnce 'afterRender', @, 'setupGridstackWidget'`

I haven't been able to grock what the "right way" to integrate plugins that directly manipulate the DOM with Ember, though. What am I missing?

gmsa:
+1

kdietrich:
You can use `remove_widget` with `detach_node=false` to remove a widget without changing the DOM. The counterpart `make_widget` allows you to convert an existing node into a widget without gridstack creating another DOM node.

jhpedemonte:
`remove_widget` isn't enough. Calling `destroy` on the Gridstack object does cleanup for the whole grid, not just an individual widget.  Looking at the code, it removes a "resize" handler and style rules (as well as removing the container from the DOM, which is what we want to avoid in this issue).

radiolips:
We can add a parameter for `destroy` to determine whether or not to remove from the DOM.

radiolips:
This feature has been merged in. `destroy` now takes an optional `detachGrid` method. Pass false and I believe your problem will be solved. Let me know if you have any further problems.

PR #346, opened 2016-02-22
Feature/216/destroy param

radiolips:
#216


src/gridstack.js
```diff
@@ -1041,10 +1041,14 @@
         this._updateContainerHeight();
     };
 
-    GridStack.prototype.destroy = function() {
+    GridStack.prototype.destroy = function(detachGrid) {
         $(window).off('resize', this.onResizeHandler);
         this.disable();
-        this.container.remove();
+        if (typeof detachGrid != 'undefined' && !detachGrid) {
+            this.removeAll(true);
```
jhpedemonte: Shouldn't this be `false`? Otherwise, when `detachGrid === false`, we won't remove the grid container, but we will remove all of the widgets.

radiolips: You're absolutely correct. This should be `this.removeAll(false)`. We're in the process of adding in unit tests, and clearly have not added this yet. Apologies for the mixup - I'll fix this and have it pushed out momentarily. Thanks for finding and reporting this bug - saved someone lots of frustration down the road. I hope this didn't cause you much issue.

radiolips: This has been updated. Thank you again for finding this!

jhpedemonte: Sweet! Thanks for the quick turnaround.

PR #360, opened 2016-02-24
Bugfix/detach grid

radiolips:
Fixes issue in new code found by @jhpedemonte (thanks!).


jhpedemonte:
This isn't fully working with latest code. Nodes are still deleted from DOM when doing a `destroy(false)`.

It works fine with commit 9c01f02 (if I change `removeAll(!0)` to `removeAll(!1)` in *gridstack.min.js*) but fails with commit 85740de (& master). Trying to find out what changed in *gridstack.min.js*.

jhpedemonte:
Hmm, wait, I may have the commits wrong. But definitely still an issue on master. Just trying to find out where things went wrong.

jhpedemonte:
Seems to be caused by commit a33f046fa6b05ad2e42f4b3eaf2d0921504f2b4b (after manually patching *gridstack.min.js* to `removeAll(!1)`). Any idea why these changes would still delete nodes when calling `destroy(false)`?

radiolips:
Okay, give it one more try. We had found a silly bug a few days ago when we were doing our unit tests. Unfortunately, it turns out that fixing that bug created a new one. I have fixed that. Thanks for your patience and thank you very much for tracking it down to that particular commit. A slight bit of a doozy, but now it should work.

For reference, `removeWidget` with `detachNode=false` was also broken. Both `destroy` as well as `removeWidget` should now work properly with `detach` false.

PR #361, opened 2016-02-24
Bugfix/detach grid 2

radiolips:
Fix one bug and another pops up. This fixes issue where passing detachNode=false to removeWidget would still cause it to be removed from the DOM.


jhpedemonte:
Thanks! This is working as expected now.

jhpedemonte:
Well, found another issue!

In our [project](https://github.com/jupyter-incubator/dashboards), when we hide a cell, we call `removeWidget(w, false)`. Because we pass `false`, we don't call `_notify`, which means we don't call the `onchange` callback [here](https://github.com/troolee/gridstack.js/blob/d521ef53301df71f10390278b486e2049a1c50fc/src/gridstack.js#L575).  That's what resets the position of widgets after one has been removed.

This used to work for our project fine with Gridstack 0.2.4. But now, any widgets below the one we deleted don't get moved up automatically (everything else, including container height, seems to still be set correctly).

How should we handle this? Is there still a defect in Gridstack code, or do we need to call another function/event in order to get things working as expected?

[Our code that handles "hiding" (i.e. removing) widgets is [here](https://github.com/jupyter-incubator/dashboards/blob/649ef611a0e01f6badd8f90407009bd596174338/jupyter_dashboards/nbextension/notebook/dashboard-view/dashboard.js#L490).]

jhpedemonte:
I can workaround this by calling `batchUpdate()` before `removeWidget(w, false)`, then calling `commit()` inside of the "change" event listener. That makes things work as before, but I don't think that's a good solution.

radiolips:
I can update `_notify` to take in an additional parameter, but it's going to need to be a bigger change and will require updating our unit testing (`_notify` is unit tested!). I'll try to get to it today, but it may have to wait until tomorrow. I will certainly fix it, though. Thanks for being a trooper through this! Cascading bug fix->new bug has been a pain for you, I'm sure. We should have all of the unit tests written in the next couple weeks to avoid this in the future.

jhpedemonte:
No worries. Like I said, we already have this working with v0.2.4, albeit by monkey patching the `destroy` method. I also don't need this ASAP, so need to hurry.

PR #367, opened 2016-02-26
Allow to specify if widgets are resized when calling setGridWidth.

pmoradei:
I added a argument 'doNotPropagate' to the setGridWidth. If this argument is explicitly passed as true the inner widgets will not be resized.

And In the setGridWidth method i also updated the GridstackEngine 'width' attribute in addition to the opts.width. Il you do not do it you will not be able to move/resize widgets to the the new space generated if the parent grid has been expanded.


radiolips:
Thanks for the feature/fix, @pmoradei! I think you may have changed the indentation from spacing to tabs. Could you change this back? Unfortunately, the diffs are showing that every single line has been changed.

We are strictly following a particular style guide now. If you do `npm install` in the gridstack directory you can run `jscs gridstack.js`. You may need to install `jscs` globally, not sure. Give that a try and make sure it's passing - As you can see from our automatic TravisCI test, it isn't currently passing. Get these things fixed and I'd love to take a look at your code and hopefully merge it in!

PR #368, opened 2016-02-26
Allow to specify if widgets are resized when calling setGridWidth

pmoradei:
I added a argument 'doNotPropagate' to the setGridWidth. If this argument is explicitly passed as true the inner widgets will not be resized.

And In the setGridWidth method i also updated the GridstackEngine 'width' property in addition to the opts.width. If you do not do it there will be two problems  :
 
-  you will NOT be able to move/resize widgets to the the new space generated if the parent grid has been expanded.
-  you will be able to resize/move a widget beyond the edge of its parent grid the grid's dimensions has been decreased.

Because in the function dragOrResize(_prepareElement) self.grid.width its used to determine if a widget is dragged outside of its parent grid.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1417,10 +1417,13 @@
         this.grid.commit();
     };
 
-    GridStack.prototype.setGridWidth = function(gridWidth) {
+    GridStack.prototype.setGridWidth = function(gridWidth,doNotPropagate) {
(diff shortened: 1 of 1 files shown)


radiolips:
Hi, @pmoradei! As I mentioned in the other PR, we really can't merge a change set that changes every line of code. Could you verify that what you're pushing are only your intended line changes?

PR #370, opened 2016-02-28
Allow negative numbers to be passed to parseHeight function

krema:
It should be possible to call for example the verticalMargin method with negative numbers.
     
    verticalMargin('-5px')

Instead of bypassing the method

    grid.opts.verticalMarginUnit = 'px';
    grid.opts.verticalMargin = '-5';
    grid._updateStyles();

### DIFF
--- a/dist/gridstack.js
+++ b/dist/gridstack.js
@@ -99,7 +99,7 @@
             var height = val;
             var heightUnit = 'px';
             if (height && _.isString(height)) {
-                var match = height.match(/^([0-9]*\.[0-9]+|[0-9]+)(px|em|rem|vh|vw)?$/);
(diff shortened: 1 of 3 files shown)


krema:
Sorry about the failed checks. I will update the pull request tonight.

radiolips:
What's the scenario when you'd want a negative vertical margin?

krema:
Let's say you have set borders for each "grid-stack-item-content" to 10px and want to display each grid item seamlessly without displaying border twice then you need to decrease the space between the items for one border width. A verticalMargin('0') is not enough to display the items seamlessly.

verticalMargin('0'):

verticalmargin0

verticalMargin('-10px'):

verticalmargin_minus6

troolee:
Please update unit tests for `parseHeight` function to add this case

PR #371, opened 2016-02-29
Add a parameter to setGridWidth method to allow not to propagate resi…

pmoradei:
Add a parameter to setGridWidth method to allow not to propagate resizing to widgets. 

I also update the this.grid.width property in the setGridWidth function. If you do not do that the GridStackEngine will not be aware of the new width and you will have 2 problems 

- you will not be able to move widget to the newly created space if the grid has been expanded.
- you will be able to to move widget beyond the grid's limit if the grid has been shrinked.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -413,11 +413,12 @@ Toggle the grid animation state.  Toggles the `grid-stack-animate` class.
 
 - `doAnimate` - if `true` the grid will animate.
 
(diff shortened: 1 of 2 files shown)


troolee:
Please update doc file `doc/README.md` with the new parameter

pmoradei:
Done.

troolee:
Thanks!

PR #421, opened 2016-04-12
remove div.ui-resizable-handle if exist from gridstack

eagoo:


eagoo:
1.auto cellHeight fix
2.tow gridstack acceptWidgets bugs fix
3.widget can not resize after drag between two grid bug fix
4.drag between tow grid,two placeholder fix

Issue #433, opened 2016-04-23
Locked items move when dropping an external item


aletorrado:
Locked items work OK while dragging in the same GridStack instance, but when I drag and drop an item from another GridStack instance, or a plain jQuery draggable object, GridStack doesn't seem to recognise any of the locked items.

aletorrado:
#434

wrenn:
Any solution for this yet?

radiolips:
@aletorrado @wrenn I'll try to take a look at this today. If I'm unable, it won't make `v0.3.0`, which will be released today, but I'll try to get this in today!

macgyver76:
Any updates on this?

radiolips:
When dragging from one grid to another, while still dragging, the held widget will only move unlocked widgets in released `v0.4.0`.

PR #434, opened 2016-04-23
Some fixes when dropping external items.

aletorrado:
Old getCellFromPixel call used ui.offset, which has the position of the upper left corner of the dragged item, but we need to track the mouse position.


KeithScheiwiller:
Is this still being reviewed? #433 causes lots of headaches so I'm currently using this patched version instead of upstream. Seems to have no side-effects.

radiolips:
@KeithScheiwiller It's in the develop branch now.

KeithScheiwiller:
@radiolips Great, thanks!

PR #447, opened 2016-05-12
Trigger event after update finished

rharriso:
there is currently no event that fires after the endUpdate method is run.

This is helpful for detecting resizes and moves after initial creation.


rharriso:
@troolee any interest in this?

PR #448, opened 2016-05-18
bugfix: include vertical margin with height calculations

rharriso:


src/gridstack.js
```diff
@@ -1048,8 +1048,8 @@
             node._beforeDragX = node.x;
             node._beforeDragY = node.y;
 
-            el.resizable('option', 'minWidth', cellWidth * (node.minWidth || 1));
-            el.resizable('option', 'minHeight', strictCellHeight * (node.minHeight || 1));
+            el.resizable('option', 'minWidth', Math.ceil(cellWidth * (node.minWidth || 1)));
+            el.resizable('option', 'minHeight', Math.ceil(Math.strictCellHeight * (node.minHeight || 1)));
```
rharriso: Rounding as late as possible means losing as little info as possible with longer grids

src/gridstack.js
```diff
@@ -1496,8 +1496,8 @@
         var relativeLeft = position.left - containerPos.left;
         var relativeTop = position.top - containerPos.top;
 
-        var columnWidth = Math.floor(this.container.width() / this.opts.width);
-        var rowHeight = Math.floor(this.container.height() / parseInt(this.container.attr('data-gs-current-height')));
+        var columnWidth = this.container.width() / this.opts.width;
+        var rowHeight = this.opts.cellHeight + this.opts.verticalMargin;
```
rharriso: here again, rounding too early causes problems with the calculations on longer grids

Issue #460, opened 2016-06-02
Grid Height not updated when added new item using cellHeight = 'auto'


Nyrris:
The situation is that I have an empty grid, and when adding a new item to it. The grid height is not updated. When I looked add the function _updateContainerHeight in the library there I found that the height is calculated with the given cellHeight, however this is in my case 'NaN' and can not be calculated.

See the function below:
```
GridStack.prototype._updateContainerHeight = function() {
        if (this.grid._updateCounter) {
            return;
        }
        var height = this.grid.getGridHeight();
        this.container.attr('data-gs-current-height', height);
        if (!this.opts.cellHeight) {
            return ;
        }
        if (!this.opts.verticalMargin) {
            this.container.css('height', (height * (this.opts.cellHeight)) + this.opts.cellHeightUnit);
        } else if (this.opts.cellHeightUnit === this.opts.verticalMarginUnit) {
            this.container.css('height', (height * (this.opts.cellHeight + this.opts.verticalMargin) -
                this.opts.verticalMargin) + this.opts.cellHeightUnit);
        } else {
            this.container.css('height', 'calc(' + ((height * (this.opts.cellHeight)) + this.opts.cellHeightUnit) +
                ' + ' + ((height * (this.opts.verticalMargin - 1)) + this.opts.verticalMarginUnit) + ')');
        }
    };
```

Nyrris:
After some research I come to this solution, the cellHeight is set to the cellWidth, however the width is calculated from the first item which in this case doens't excists.

```
GridStack.prototype.cellWidth = function() {
        var o = this.container.children('.' + this.opts.itemClass).first();
        return Math.ceil(o.outerWidth() / parseInt(o.attr('data-gs-width'), 10));
    };
```

The cellWidth schould be calculated from the container class and number of columns like:
```
GridStack.prototype.cellWidth = function() {
        var o = $(this.container);
        return  Math.ceil(o.outerWidth() / o.attr('data-gs-width'));
    };
```

This will solve my issue.

radiolips:
Hm. I tested this and it works okay for me. I can see how this might be an issue, but it does work for me. Could you setup a jsfiddle and I will take a closer look?

Nyrris:
Sorry for my late response, I did tried to reproduce the problem in a fiddle but there it seems to work correctly. To bad I can not copy my code base into a fiddle, I will try out the new version and let you know if the problem still exists.

PR #461, opened 2016-06-02
#460 Grid Height not updated with cellHeight=auto

Nyrris:
#460 Fixes grid height not updated when added new item using cellHeight = 'auto'

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1488,8 +1488,8 @@
     };
 
     GridStack.prototype.cellWidth = function() {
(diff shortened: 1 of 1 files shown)


Fathardie:
Would love this fix in 0.26 if possible!

radiolips:
`v0.2.6` will hopefully go out tonight or tomorrow. We will take a look at this as soon as we can. This fix or another should go in `v0.2.7`. We are hard at work with fixes and new features at the moment, so `v0.2.7` shouldn't be far away.

PR #490, opened 2016-08-04
fixed rounding cellWidth twice

LomoSoftGmbH:
When using more cols > 12, in my case 40, the doubled rounding forces an unprecise dragging/dropping of widgets


troolee:
Hi @LomoSoftGmbH,

thanks for PR, but I didn't get it. Code is the same, you just moved `Math.ceil` inside the call. Did you  want to get smth like:

`Math.ceil(cellWidth * (node.minWidth || 1))`
`..................^^^....................^^^`

PR #500, opened 2016-08-15
Add jQuery UI droppable as requirement

nikcub:
Allow bespoke builds of jQuery UI by adding droppable as a module requirement


radiolips:
Thanks, @nikcub ! We must have missed this when we added dropping to other grids.

nikcub:
Not a problem! 

I didn't notice the requirement either until we did a custom production build that attempted to slim things up

PR #501, opened 2016-08-15
Add grunt-cli as dev dependancy

nikcub:


radiolips:
I only just read https://nodejs.org/en/blog/npm/npm-1-0-global-vs-local-installation/ - very interesting!

nikcub:
It is a good idea, IMO - i've been keeping my global with nothing but standard lib for a while now

PR #507, opened 2016-08-17
Now use top and bottom of widget instead of just center height when c…

Fathardie:
…alculating remove. Fixes the removal of widgets when top leaves grid


radiolips:
Thanks! This is actually resolved in `v0.3.0`. We went a slightly different route, keeping the `y` as it was, and added in a check for float to be more reliable.

PR #551, opened 2016-11-03
trigger only one event when calling removeAll()

sthuck:
Hi guys,

Currently when calling removeAll(), the 'removed' event get triggered for every item on the grid, and it passes only one item to the event handler. I think a more logical behavior is triggering the event only once. Otherwise, there isn't really a point for '_removedNodes' on GridStackEngine. Hope you'll find this change needed.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1288,8 +1288,9 @@
         return this.grid.canBePlacedWithRespectToHeight(node);
     };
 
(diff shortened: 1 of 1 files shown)


radiolips:
I'm not sure if this is the right way to go or not. I think about this every so often. It haunts me. I can understand both sides.

adumesny:
thank you for the suggestion - I ended up checking in a better solution #1097

PR #587, opened 2017-01-11
Avoid collision check of node with itself. Fixes #586

### LINKED ISSUE #586: Widgets with a height larger than 1 do not return to their original position
When I drag widget A to the position of widget B (height > 1), widget B moves aside as expected. But when I decide to put widget A to a different position (while I am still dragging it), widget B will not return to its original position.

You can find a small example of my problem here: https://justuslauten.github.io/gridstack

Drag one of the grey widgets to the position of the red widget, and back. The red widget will not return to its original position.

hairyheron:
`_packNodes` did not work correctly in floating mode for nodes of height > 1. When trying to move such a node upwards, the repositioned node intersected with its previous position. Fix this by only checking for collisions with preceding nodes instead of all nodes.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -233,6 +233,7 @@
                 var newY = n.y;
                 while (newY >= n._origY) {
                     var collisionNode = _.chain(this.nodes)
+                        .take(i)
                         .find(_.bind(Utils._didCollide, {n: n, newY: newY}))
                         .value();
 


adumesny:
you can use https://github.com/troolee/gridstack.js/blob/develop/demo/two.html to test your fix using grey box (float grid) and it fixes the issue.

PR #622, opened 2017-04-03
Added trigger('remove') to notify .on("remove")

RobertoBiundo:
Added trigger('remove') to notify .on("remove")
This is causing .on("remove") not getting notified in some cases.
jQuery doesn't trigger the remove callback when .remove() its called


radiolips:
Thanks for this! Unfortunately, we need to be able to silently remove nodes in some cases. I'm meanwhile putting in a fix that this would have solved. Thank you, though!

PR #641, opened 2017-04-23
Workaround for undefined width/height during drag/drop

swarnat:
Hy,

Working with this great JS Library, throw an exception **_placeholder.attr(..).attr(..).attr(..).attr(..) is not a function_**.
During a Debug session, I recognize, that during Drag&Drop the Width and Height of Placeholder is "undefined". 

I tried to enable the width calculation, but this don't work, because Resizing is not initialized. (Correct)

If I look at this part of code, I understand why it don't work, because width is undefined and the attr(...)  function with _undefined_ second parameter returns the value of attribute and not the element.

I don't fully understand, when and why **node._temporaryRemoved** is true AND width is undefined, because I think if this happens, the placeholder would be initialized with wrong width. But everything works like it should in my environment.

Maybe this PR is only a workaround for another issue. I will further check the function.

Regards,

Stefan

PS: I use jQuery 1.x, because your library will work within another software, which force this version. 
It works without problems.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1088,10 +1088,13 @@
                         self.grid.addNode(node);
                         self.placeholder
                             .attr('data-gs-x', x)
-                            .attr('data-gs-y', y)
-                            .attr('data-gs-width', width)
-                            .attr('data-gs-height', height)
-                            .show();
+                            .attr('data-gs-y', y);
+                        if(typeof width !== 'undefined' && typeof height !== 'undefined') {
+                            self.placeholder
+                                .attr('data-gs-width', width)
+                                .attr('data-gs-height', height);
+                        }
+                        self.placeholder.show();
                         self.container.append(self.placeholder);
                         node.el = self.placeholder;
                         node._temporaryRemoved = false;


radiolips:
@swarnat I'm able to drag and drop without problem. Could you provide a jsfiddle so I can better understand what this fixes?

swarnat:
Hy,
Thanks for response and please forget it.
I invest some time and recognize this behaviour was changed with jQuery 1.7.2 (https://bugs.jquery.com/ticket/9427)
Because you request version 3 of jquery, the error is not a big problem.
Unfortunately the CRM I use, still use jQuery 1.7.0 and behave different in this situation.

I create this JSFiddle with jQuery 1.7.1:
https://jsfiddle.net/8pdk54p2/1/

You could reconstruct the "bug", if you Drag the "Content B" Item over the text "Headline", so the Placeholder will be temporarily removed. It occurs during recreation of Placeholder and destroy the complete grid.

I will use your library with this modification and during next major update of CRM, which update jQuery to 1.11, it is getting cleared.

Stefan

PR #672, opened 2017-05-24
bug fix: keep this._styles._max increasing

datou3600:
After calling `this._initStyles()`, the condition `this._styles._max !== 0` will never be true. And then a smaller `maxHeight` will be able to assign to `this._styles._max`, and this will cause the missing of some style rules.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -947,12 +947,12 @@
         if (typeof maxHeight == 'undefined') {
             maxHeight = this._styles._max;
         }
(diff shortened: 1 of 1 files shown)


radiolips:
@datou3600 could you provide an example of this breaking?

datou3600:
@radiolips glad to see that you had found an example. This situation usually happens when an element was in the bottom of the DOM, and it was dragged to the top position.

PR #684, opened 2017-06-08
Fixed incorrect attribute usage in documentation

jhadenfeldt:
data-gs-no-resize, data-gs-no-move, data-gs-auto-position and data-gs-locked need to be set to "true" (or "1" or "yes") in order to take effect. Simply setting the attributes is not sufficient.


radiolips:
@jhadenfeldt Could you actually move the `=true` part into the sentence that explains it? Something like "Set to `true`, `yes`, or `1`, to ...".

radiolips:
Closing for inactivity, but if you have a chance to update this before it otherwise gets updated, please put in a new PR!

PR #697, opened 2017-06-22
Fix verticalMargin setter.

jonapgar:
Fixes bug where verticalMargin is string (eg: 12px) and has same numeric value as opt.height.


radiolips:
@jonapgar Can you provide a fiddle or some environment where I can see what exactly this bug is?

radiolips:
I'm closing this because there wasn't an example provided, but if you can show me where this is fixing something in a new PR, I will happily merge!

PR #732, opened 2017-07-31
Use v1 API options in demo page

demedos:
Options `cell_height` and `vertical_margin` are deprecated as of v0.2.5 and have been replaced with `cellHeight` and `verticalMargin`. They will be **completely** removed in v1.0.


radiolips:
Thanks, @demedos !

PR #733, opened 2017-08-02
Fixed a bug where multiple resize handles are added to a node.

veselfil:
When moving nodes between multiple grids, the resize handle was added multiple times to the node which caused the resize function to not work properly.

To replicate this problem, just open the two grid demo, try moving a node from one grid to the other and then try resizing it.


radiolips:
With the latest merged commit, I don't think this is any longer necessary. Please reopen and ping me if this is not the case.

adumesny:
this should have fixed it https://github.com/troolee/gridstack.js/pull/769

PR #737, opened 2017-08-13
add staticHeight option

silverwind:
This is useful with the `float` option when one wants the grid to not automatically size down when moving items. Would have liked to have a similar option for width, but from the looks of it, width already never changes.


silverwind:
I fear I have to close it, it's not really complete. I have a branch that fixes some of the issues, but it's not really complete (one issue is that the collision check can place items outside the fixed height). Maybe I'll get this in a more complete state later.

PR #747, opened 2017-08-22
links to docs

luckylooke:


doc/README.md
```diff
@@ -90,12 +90,12 @@ gridstack.js API
 - `verticalMargin` - vertical gap size (default: `20`). Can be:
  - an integer (px)
  - a string (ex: '2em', '20px', '2rem')
-- `width` - amount of columns (default: `12`)
+- `width` - amount of columns (default: `12`). Setting non-default value must be supported by equivalent change in CSS, [see docs here](https://github.com/troolee/gridstack.js#change-grid-width).
```
adumesny: no wonder why non default 12 columns didn't work well when I tried... having to change CSS is less than ideal for me as I'm letting user eventually configure grids data like # of columns - this for custom dasboarding, and creating coumpound widgets (tight griditem inside a smaller nested grid)

PR #748, opened 2017-08-22
Allow percentage as a valid unit for height

luckylooke:
percentage is also valid css unit for height


radiolips:
This seems like it would cause problems in some places. No? I know that height in percentage is really funky across different browsers.

luckylooke:
Maybe, but should be let on user to decide, not all percentage cases are so funky ;) We are using flex and we did not catch any problems with percentage heights yet. :)

adumesny:
actually % height makes perfect sense for nested grids where as fixed pixel doesn't!
I mean if I have a parent cell (which user can resize at will) then children probably have to take entire height and be % based. I doubt the code would allow you to add/remove row based on drag gesture though...

adumesny:
using #1093 as I wasn't able to merge it for some reason...

PR #765, opened 2017-09-06
Preserve inline styles when moving/cloning items

silverwind:
Instead of removing all inline styles on moved/cloned items via `removeAttr('style')`, it's better to remove just the positioning styles that were previously added.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -106,6 +106,13 @@
                 height = parseFloat(match[1]);
             }
             return {height: height, unit: heightUnit};
+        },
+
+        removePositioningStyles: function(el) {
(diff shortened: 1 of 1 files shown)


silverwind:
Demo of this issue: https://jsfiddle.net/silverwind/55ssnhzk/

Note how the red background color is lost after dragging a item.

radiolips:
@silverwind thanks for the example! I'm merging, but does it also need to unset width/height?

silverwind:
Yes, I think unsetting width and height as well might be necessary for resizable items. Can you add those?

radiolips:
(I added them. They were definitely needed)

PR #775, opened 2017-09-13
Min height fix

adumesny:
use verticalMargin when calculating min height in rows

Testing:
1. run two.html and check that yellow grid has height of 300px
2. modify two.html to have
#grid1 {
  background: lightgoldenrodyellow;
   min-height: 300px;
}
3. rerun and notice grid1 now has 380px actual height (passed last item) !
4. adding removing items will shrink grow, but not to the correct min 300px size.


package.json
```diff
@@ -9,7 +9,8 @@
   },
   "scripts": {
     "build": "grunt ; doctoc ./README.md ; doctoc ./doc/README.md ; doctoc ./doc/CHANGES.md",
-    "test": "grunt lint && karma start karma.conf.js"
+    "test": "grunt lint && karma start karma.conf.js",
+    "lint": "grunt lint"
```
adumesny: added an explicit lint (makes it easier to check before commit) as for some reason it's not auto linting in Visual Studio code yet...

src/gridstack.js
```diff
@@ -1033,9 +1033,13 @@
             return;
         }
         var height = this.grid.getGridHeight();
```
adumesny: real change - ingore the /dist files as they will be removed yet (merge hell)

src/gridstack.js
```diff
@@ -1033,9 +1033,13 @@
...
-        var minHeight = parseInt(this.container.css('min-height')) / this.cellHeight();
-        if (height < minHeight) {
-            height = minHeight;
+        // check for css min height. Each row is cellHeight + verticalMargin, until last one which has no margin below
```
adumesny: real change - ignore the /dist files as they will be removed yet (merge hell)

adumesny:
introduced by https://github.com/gridstack/gridstack.js/pull/767
fixes https://github.com/gridstack/gridstack.js/issues/628

PR #777, opened 2017-09-13
remove /dist

### LINKED ISSUE #776: [internal] remove `dist/` from versioning, add needed release steps
Versioning `dist` adds unneeded changes for contributor PRs. Ideally `dist` should not be versioned and instead only as part of `npm publish` or a release process.
# todo 
- [ ] add to .gitignore
- [ ] modify `Gruntfile.js` with publish task, incorporate into other steps

adumesny:
remove /dist folder and instead generate the output files during build process, so there is no conflict on merge or issue being out of sync.

use 'build' target to generate before publishing npm (which I didn't see included)

fix #776


package.json
```diff
@@ -8,9 +8,10 @@
     "url": "git+https://github.com/gridstack/gridstack.js.git"
   },
   "scripts": {
-    "build": "grunt ; doctoc ./README.md ; doctoc ./doc/README.md ; doctoc ./doc/CHANGES.md",
+    "build": "mkdir dist && grunt && doctoc ./README.md && doctoc ./doc/README.md && doctoc ./doc/CHANGES.md",
```
adumesny: && are needed to run on windows instead of ;

efx: `mkdir dist` will fail if mkdir exists. You can pass `mkdir -p` to only create it if it does not exist

efx: looks like `-p` is only supported on *nix versions of the utility

PR #784, opened 2017-09-17
Update documentation and force x/y coords to be stores as ints.

radiolips:
### Description
Piggyback on idea from https://github.com/gridstack/gridstack.js/pull/648 - when x and y coordinates for widgets are stored, run `parseInt(num, 10)` to ensure that they're treated as integers when sorting.

### Checklist

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1272,8 +1272,8 @@
 
         el.addClass(this.opts.itemClass);
         var node = self.grid.addNode({
-            x: el.attr('data-gs-x'),
-            y: el.attr('data-gs-y'),
+            x: parseInt(el.attr('data-gs-x'), 10),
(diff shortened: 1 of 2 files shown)


src/gridstack.js
```diff
@@ -1272,8 +1272,8 @@
 
         el.addClass(this.opts.itemClass);
         var node = self.grid.addNode({
-            x: el.attr('data-gs-x'),
-            y: el.attr('data-gs-y'),
+            x: parseInt(el.attr('data-gs-x'), 10),
+            y: parseInt(el.attr('data-gs-y'), 10),
             width: el.attr('data-gs-width'),
```
adumesny: is there a reason you are not doing width/height/maxwidth/etc... as int as well ? el.attr() returning string should be checked throughout code IMO.

PR #786, opened 2017-09-18
fixed cellHeight() check for no-op return

adumesny:
the code that checked for cell height and unit being the same was incorrect.

Testing Done:  calling this should be no-op, but wasn't
this.grid.cellHeight( this.grid.cellHeight() );

code right below uses correct names
this.opts.cellHeightUnit = heightData.unit;
this.opts.cellHeight = heightData.height;

P.S. this is why Typescript is good - you don't do those stupid JS typos....

### DIFF
--- a/spec/gridstack-spec.js
+++ b/spec/gridstack-spec.js
@@ -5,16 +5,16 @@ describe('gridstack', function() {
     var w;
     var gridstackHTML =
         '<div class="grid-stack">' +
+        '  <div class="grid-stack-item"' +
+        '  data-gs-x="0" data-gs-y="0"' +
+        '  data-gs-width="4" data-gs-height="2">' +
+        '    <div class="grid-stack-item-content"></div>' +
+        '  </div>' +
         '   <div class="grid-stack-item"' +
-        '   data-gs-x="0" data-gs-y="0"' +
-        '   data-gs-width="4" data-gs-height="2">' +
-        '       <div class="grid-stack-item-content"></div>' +
+        '   data-gs-x="4" data-gs-y="0"' +
+        '   data-gs-width="4" data-gs-height="4">' +
+        '     <div class="grid-stack-item-content"></div>' +
         '   </div>' +
-        '    <div class="grid-stack-item"' +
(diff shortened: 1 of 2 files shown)


spec/gridstack-spec.js
```diff
@@ -5,16 +5,16 @@ describe('gridstack', function() {
     var w;
     var gridstackHTML =
         '<div class="grid-stack">' +
```
adumesny: those are just formatting 2 spaces change. ignore.

spec/gridstack-spec.js
```diff
@@ -244,6 +244,38 @@ describe('gridstack', function() {
...
+                verticalMargin: 10,
+                width: 10
+            };
+            $('.grid-stack').gridstack(options);
+            var grid = $('.grid-stack').data('gridstack');
+            grid.cellHeight( 120 );
+            expect(grid.cellHeight()).toBe(120);
+        });
+    });
+
```
adumesny: added simple test above to call cellHeight() after the fact to get more coverage. I don't know how to test to make sure grid changes height when increasing row height (80->120) which would fail to work correctly in the nested case (but might be ok otherwise).

spec/gridstack-spec.js
```diff
@@ -838,9 +870,9 @@ describe('gridstack', function() {
             $('.grid-stack').gridstack(options);
             var grid = $('.grid-stack').data('gridstack');
             var widgetHTML =
-                '   <div class="grid-stack-item">' +
```
adumesny: again indentation fixes. ignore

src/gridstack.js
```diff
@@ -1610,7 +1610,7 @@
         }
         var heightData = Utils.parseHeight(val);
 
-        if (this.opts.cellHeightUnit === heightData.heightUnit && this.opts.height === heightData.height) {
+        if (this.opts.cellHeightUnit === heightData.unit && this.opts.cellHeight === heightData.height) {
```
adumesny: code fix to use actual vars

Issue #757, opened 2017-09-03
How to get coordinates, which actually changed


vadus1:
When I change the position of the widget, such as drag two widgets. Using the event 'change' to get these updated widgets. 
I make a condition, which checks the changed position.
so . `if (stack_item.data('gs-x') != items[i].x || stack_item.data('gs-y') != items[i].y) {`
I reload the page. Change position of the widget, and nothing happens. Then again change the position of the same widget and I gets position.

why? How to get coordinates, which actually changed?
```javascript
$(document).on('change', '.grid-stack', function(event, items) {
  var updated_widgets = {};
  for (var i = 0; i < items.length; i++) {
    var widget = items[i].el.find('.dashboard-widget');
    var stack_item = widget.parent();
    // Let's see if cordinates actually changed
    // (this change event reports changes even when nothing changed)
    if (stack_item.data('gs-x') != items[i].x || stack_item.data('gs-y') != items[i].y) {
      updated_widgets[widget.data('widget-id')] = {
        'x': items[i].x,
        'y': items[i].y
      }
    }
  }
```

radiolips:
As seen in this fiddle https://jsfiddle.net/9tga7oam/39/ , both widgets are returned for every resize/move rather than just the affected widget.

kstan79:
I think this callback check DOM variable which is not immediately effect, i'm overcome this by delay 50ms:

$('.grid-stack').on('resizestop', function(event, ui) 
				{
                	var element = event.target;    				
                	var uuid=$(element).find('.bi2_element_body').attr('id')
                	// need wait a while
                	setTimeout(function()
                		{
                			console.log($(element).attr('data-gs-width')); 
                		},50);
                	
                });

adumesny:
closing as change callback happens only to what actually changed, and after the dom gs- attributes have been updated.

PR #789, opened 2017-09-19
#757 Fixed returning of an array of actually changed widgets

vadus1:


radiolips:
@vadus1 What's this fix?

adumesny:
Note: this appears to be a better fix (guessing when items are moved back to original spots current code would incorrectly mark them as dirty) **ASSUMING**  `beginUpdate() / endUpdate()` are called only once and for each cases - which I have not tested - appears user driven gesture `onDrag, onStartMoving` do, but not sure about API to add/remove items.... I know `float(val)` I recently added **does not**.

adumesny:
this is NOT the right fix, because if you temporally move other objects out while dragging one, they will not return to original position (_dirt from original will be FALSE, so `new GridStackEngine() onChange()` callback will not be called to do a `self._updateStyles(maxHeight + 10);` on line 784

PR #809, opened 2017-10-25
Fix for changing grid width

nulen:
Fixes problem with cell positioning when grid width is changing. Now fields are positioned as intended (or I hope so).

### Description
Updated setGridWidth method to update nodes widths only after grid width is changed.
Updated _updateNodeWidths method to prevent changing list during looping throgh it.


radiolips:
@nulen Could you provide an example of this issue?

adumesny:
thank you for the suggestion - got me intrigued in fixing it right (1 column which is trivial always bugged me being wrong). fixed by #1098

PR #810, opened 2017-10-25
Fix for grids with many columns

nulen:
There was a problem with positioning dragged widgets if there were larger number of columns and cumulated diff between real value and rounded one. (e.g. cellWidth is calculated to be 19.5px and there are 30 columns. Further we go, larger the mistake with positioning widgets)

### Description
Updated cellWidth method to not round cellWidth value. It was causing problem for grids with larger number of columns and uneven cell width calculated.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1623,7 +1623,7 @@
     };
 
     GridStack.prototype.cellWidth = function() {
-        return Math.round(this.container.outerWidth() / this.opts.width);
+        return this.container.outerWidth() / this.opts.width;
     };
 
     GridStack.prototype.getCellFromPixel = function(position, useOffset) {


radiolips:
This is an exact value and not a percentage. I'd imagine in every case you'd want this to return the rounded value. Why is this causing problems?

adumesny:
if we use cellWidth() * n, then the rounding will mess things up. I agree with him, taking a look.

adumesny:
Created a sample for #1053 with 30 columns, but not seeing issues in that PR (issue rounding cellWidth() with large # of columns). I was able to move and resize and things lined up as expected even when total width / 30 was close to .5 off
@nulen please update with info.

adumesny:
@nulen closing for now I don't see the need for this 'fix' - please re-open and update with more info (see my comment and sample app)

adumesny:
@nulen closing for now - please re-open with comments if you have steps to reproduce issue (see my demo app). thank you.

bpwalker:
A breaking example is use a large number of columns, say 100, and try and move a widget that has a larger x position (far right of the grid). In dragOrResize, ui.position.left / cellWidth will result in a huge error that will result in the widget moving significantly even if you've barely touched it.

adumesny:
thank you, I ended up checking in the fix for 2.x

PR #822, opened 2017-11-29
Improved item placement in _updateNodeWidths()

dswitzer:
### Description

I noticed there are several bugs in the `GridStack.prototype._updateNodeWidths` that can lead to elements not being placed correctly when the grid size changes. This causes issues when trying to use Gridstack in a responsive design.

Here's a summary of what my patch changes:

1) Not all the items are guaranteed to be processed. This is because the update() method can change the this.grid.nodes array. This can lead to items being ignored. The code now processes each item in the grid.
2) The code does not correctly confine items to the grid's width. This code adds sanity checks to make sure that an item is always placed inside the grid and that it's width does not overlap the grid's width.

### Checklist
- [ ] All tests passing (`npm test`)

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1672,10 +1672,41 @@
         this.grid._sortNodes();
         this.grid.batchUpdate();
         var node = {};
-        for (var i = 0; i < this.grid.nodes.length; i++) {
-            node = this.grid.nodes[i];
-            this.update(node.el, Math.round(node.x * newWidth / oldWidth), undefined,
-                Math.round(node.width * newWidth / oldWidth), undefined);
+
+        // We need an array of the elements to update so that we make sure that
+        // each item in the grid gets processed. We need to get a fixed list of
+        // array items, because the this.update() can shift items in the this.grid.nodes
+        // array which can lead to elements not being processed.
+        var nodes = _.map(this.grid.nodes, 'el')
+        	, columnPos, columnWidth;
+
+        for (var i = 0; i < nodes.length; i++) {
+        		// get the current live node data, in case a previous update has changed anything
+            node = this.grid.getNodeDataByDOMEl(nodes[i]);
+            /*
+             * Calculate the new column position for the item, but make sure it fits inside the
+             * grid range. If the calculated position is outside the range, then it will move
+             * it to the last column (newWidth-1).
+             */
+            columnPos = Math.min(Math.round(node.x * newWidth / oldWidth), newWidth-1);
+            // calculate the new width of the item
+            columnWidth = Math.round(node.width * newWidth / oldWidth);
+
+            // if the item is too wide, we need to adjust our calculations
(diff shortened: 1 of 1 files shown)


troolee:
Please make sure PR passes CI tests including lint.

Thank you.

dswitzer:
It looks like it's just complaining about the placement of the comma, but I can't figure out how to resolve the issue in the pull request. I can't seem to modify the file. My Google-fu must be failing me, because I'm not seeing any directions on fixing the problem.

troolee:
Just push the changes to the repository. It should update PR as well

radiolips:
Hey @dswitzer ! I think this code makes sense, but could you provide an example where making this change fixes something?

I'm very sorry to be two years late to this party! I promise that if you have a little time to address this, I will happily merge :)

dswitzer:
@radiolips,

I apologize that I didn't document the exact use cases that cause the issue. It's been a long time, so my memory is foggy on the details.

What I know is we were having issues with grid elements not always appearing withing the grid when the grid size was changing. We'd have elements that were dangling outside the grid or that were not getting shifted at all.

Looking at my changes, I think there here are the two issues I was fixing:

# The update() call could lead to an element getting updated in the this.grid.nodes array where it wouldn't get processed. For example, if position 2 gets moved into position 3, element 3 goes to position 2 and it would never get processed.
# Sometimes when resizing, an element on the right could end up too wide to actually fit within the grid size. So, you could end up with an item on the right of the row that was like 3 grid spaces wide, but there was only 2 grid spaces available.

src/gridstack.js
```diff
@@ -1672,10 +1672,41 @@
...
-        for (var i = 0; i < this.grid.nodes.length; i++) {
-            node = this.grid.nodes[i];
-            this.update(node.el, Math.round(node.x * newWidth / oldWidth), undefined,
-                Math.round(node.width * newWidth / oldWidth), undefined);
+
+        // We need an array of the elements to update so that we make sure that
+        // each item in the grid gets processed. We need to get a fixed list of
+        // array items, because the this.update() can shift items in the this.grid.nodes
+        // array which can lead to elements not being processed.
+        var nodes = _.map(this.grid.nodes, 'el')
```
adumesny: this.grid.nodes.map(n => n.el)
ES6 without lodash

adumesny:
@dswitzer thank you for making this submit!
I ended up completely re-writing `_updateNodeWidths` (now folded into setColumn()) as #1098 before seeing your changes as it was always bugging me, especially 1 column (easier one).
Your fix above ins't actually correct - if you try my column.html test case, going to 1 column (easy part) has
[2]
[1]
[0]
items are in reverse order - as as soon as you get collision, the order gets messed up, you need to re-insert in reverse and have no content to conflict...
also going back to 12 or other sizes is funky. closing with my fixes.

PR #850, opened 2018-02-14
Disallow grid overgrow

justcfx2u:
This patch disallows the grid to grow beyond the 'height' specified in the gridstack options.

Without this patch, it is trivial to grow the grid vertically where it may be undesirable to do so, assuming the height option is to be treated as a maximum size.

With this patch, the grid no longer grows beyond the maximum height specified in the gridstack options.

Video demonstration before/after patch:
https://youtu.be/ypX7Wf2G1pQ

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -425,7 +425,7 @@
             }));
 
         if (typeof clonedNode === 'undefined') {
-            return true;
+            return false;
         }
 
         clone.moveNode(clonedNode, x, y, width, height);


krilllind:
@justcfx2u You are a legend man!

radiolips:
@krilllind Did you have a chance to test this PR out?

krilllind:
@radiolips No, just saw the video, and if that works it solves a lot of problems for me.

rob101:
This commit prevents additional rows exceeding the grid height, but not if the grid is full.

radiolips:
@justcfx2u I'm closing this because of @rob101 's catch. I apologize for the long delay, but if you're still interested in putting in a new PR that also solves the case @rob101 mentioned, I'll make sure it gets merged promptly.

Issue #219, opened 2015-08-25
Can't easily drag widgets below widgets on the bottom border of a grid


derekm:
Gridster pops a few blank rows at the bottom of a grid, making it easy for your drag to scroll the window downward. When a gridstack grid has filled the window, it is impossible to drag a widget below any widget that touches the bottom border. Turning on float causes drags to add extra rows as needed. With float off, it could do the same then let the block fall upward into position.

svdb:
Just wrap your gridstack DIV inside a parent DIV, then give the parent DIV a padding-bottom >= the stackgrid row height.
So when you move a widget down you always have extra space and stackgrid will add a new row.

<div style="padding-bottom: 100px">
  <div class="grid-stack"></div>
</div>

radiolips:
@derekm did adding padding work? It should. I'm closing this issue, but please feel free to reopen it if you still have a problem.

frankadrian:
Hey @radiolips, This does not seem to work for items that are being dragged from another grid. I will try create a example for this!

frankadrian:
The behaviour can be seen here: http://troolee.github.io/gridstack.js/demo/two.html

Try dragging an item from one grid below an item in the other grid!

trongletps:
i had same issue with @frankadrian. When I set float equal false, to difficultly to drag the widget below one from the grid to the another grid.

trongletps:
Everything work well except this case but it's very important with user experience, so uncomfortable when drag and drop item from sidebar to main grid (drag item from sidebar below the other items in main grid).

stevenliebregt:
What is the status off this issue?

adumesny:
see #1094 as well

adumesny:
this might get fixed along #1001 as we will add an extra temporary row the moved/resized item is at the bottom.

adumesny:
it is now possible, though a little rough with fix to #1001 to auto-scroll

https://user-images.githubusercontent.com/3496166/103484677-93d3b580-4da5-11eb-904f-02bff4e43eba.mp4

adumesny:
should work a lot better now in next release.

we now automatically insert extra rows when dragging an item at the bottom below others to make it easier to insert below.

https://user-images.githubusercontent.com/3496166/113513223-708fe200-951d-11eb-967b-6ddd3670804e.mp4

PR #874, opened 2018-05-08
Add extra row when accepting widgets

WiktorStarczewski:
### Description
Addresses issue #219.
When accepting widgets dragged from outside, the grid will allocate one more row to allow drops at the bottom.

### Checklist

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -175,10 +175,11 @@
 
     var idSeq = 0;
 
-    var GridStackEngine = function(width, onchange, floatMode, height, items) {
+    var GridStackEngine = function(width, onchange, floatMode, height, items, extraRow) {
         this.width = width;
         this.float = floatMode || false;
         this.height = height || 0;
+        this.extraRow = extraRow || false;
 
         this.nodes = items || [];
         this.onchange = onchange || function() {};
@@ -422,7 +423,8 @@
                     return clonedNode;
                 }
                 return $.extend({}, n);
-            }));
+            }),
+            this.extraRow);
 
         if (typeof clonedNode === 'undefined') {
             return true;
@@ -454,7 +456,8 @@
             null,
             this.float,
             0,
-            _.map(this.nodes, function(n) { return $.extend({}, n); }));
+            _.map(this.nodes, function(n) { return $.extend({}, n); }),
+            this.extraRow);
         clone.addNode(node);
         return clone.getGridHeight() <= this.height;
     };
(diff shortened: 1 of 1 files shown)


Guite:
:+1:

adumesny:
we should probably combine this with the idea of having a minimum height of 1 (user settable) when all items are removed the grid collapses making it impossible to drag something back in.

In my app I hack around this by having a min CSSS height and setting a DIV with instructions (drag a new widget in) that show/hide based on events from the grid having children or not.

radiolips:
@adumesny I agree. I think this is a feature we should have, but it's probably best if we have two options (something like `extraRows: n || 0` and `minHeight: n || 1`). The use case referred to here is good, but even in a single-grid scenario, both of these things make sense. @WiktorStarczewski Thank you so much for putting in this PR! It'd be impossible to gridstack without the help of the community!

WiktorStarczewski:
@radiolips My pleasure. I think you nailed it in your comment also, since ```n``` extraRows would ease dragging down widgets causing a scroll and allowing to comfortably drop below the last one, and ```minHeight``` would allow to drop on whitespace below the bottommost widget when there is no scroll (total_widget_rows*row_height < grid_height). Thats pretty much the gripes my users have with this usecase.

src/gridstack.js
```diff
@@ -690,7 +694,7 @@
                 }
             });
             self._updateStyles(maxHeight + 10);
-        }, this.opts.float, this.opts.height);
+        }, this.opts.float, this.opts.height, null, !!this.opts.acceptWidgets);
```
adumesny: I see you only add a row if you can accept widgets, which makes sense.

src/gridstack.js
```diff
@@ -175,10 +175,11 @@
 
     var idSeq = 0;
 
-    var GridStackEngine = function(width, onchange, floatMode, height, items) {
+    var GridStackEngine = function(width, onchange, floatMode, height, items, extraRow) {
         this.width = width;
         this.float = floatMode || false;
         this.height = height || 0;
+        this.extraRow = extraRow || false;
```
adumesny: maybe should be a number instead of T/F (default to 0).

adumesny:
feels like adding a row(s) all the time is a kludge... instead the grid should grow when the user drags past the mid point of the bottom grid item and just lets you insert it below. The user heuristics ought to be enough...

adumesny:
see #1094 as well

adumesny:
I don't think this is the correct behavior - dragging to the bottom half of a widget should insert below (posted improvements) without artificially adding dead space.  Thanks for the code review though!

PR #910, opened 2018-06-15
fix for #451

### LINKED ISSUE #451: Dragging between gridstack issues: Disappearing gridstack-item, duplicate gridstack-item, resizing gridstack-item.
Hi Gridstackers.

I've found a couple of bugs regarding moving gridstack-items from one gridstack to another:
1. Disappearing gridstack-items
   
   > https://gyazo.com/dd73e460f8b931ea6cfde71065d6f2b2
2. Creates "Invisible object" / duplicate object
   
   > https://gyazo.com/b884f98970411fe77e56e3766a8ba73b
   > 
   > eagoo comes up with a fix for this inhere:
   > https://github.com/troolee/gridstack.js/issues/393
   > 
   > However that causes for gridstack-items to sometimes lock in place inside a grid: 
   > https://gyazo.com/d49221dff820cb1ed17f07c6f8e7586b
3. Unable to resize
   
   > https://gyazo.com/f73d9831d429f55bfcfecb89d94e62b9
   > 
   > eagoo comes up with a fix for this inhere:
   > https://github.com/troolee/gridstack.js/issues/416
   > 
   > Doesn't seem to have any flaws, but dont know if there's a reason why it isnt in source distribution.

Any comments, hints or fixes are highly appreciated.
- Lasse.

andrei-gheorghiu:
### Description
It seems to fix #451 but, since I'm new to Gridstack, I'd like someone with more intimate knowledge looking over proposed changes and figure out if it doesn't have any unwanted implications. 

I'm calling `cleanNodes()` inside `onEndMoving()` so when `onStartMoving()` is triggered on another gridstack container, current gridstack item doesn't get deleted. 
This change seems to trigger errors when trying to drag an item in between two different grid containers so I had to add a condition on `gridstack.js:731`.

### Checklist

### Note
For my tests, I have upped `two.html` dependencies to:
- Bootstrap (all) `3.2.0` > `3.3.7`
- jQuery `1.11.1` > `3.3.1`
- jQueryUI `1.11.0` > `1.12.1`
- lodash `4.17.0` > `4.17.10`
- Gridstack (all) `1.0.0-dev`

I haven't committed changes made to `two.html` into this pull request but I'm ready to, if deemed worthy.

### DIFF
--- a/dist/gridstack.js
+++ b/dist/gridstack.js
@@ -728,11 +728,13 @@
                         n.el.remove();
                     }
                 } else {
-                    n.el
-                        .attr('data-gs-x', n.x)
-                        .attr('data-gs-y', n.y)
-                        .attr('data-gs-width', n.width)
-                        .attr('data-gs-height', n.height);
+                    if (n.el) {
+                        n.el
+                            .attr('data-gs-x', n.x)
+                            .attr('data-gs-y', n.y)
+                            .attr('data-gs-width', n.width)
+                            .attr('data-gs-height', n.height);
+                    }
                 }
             });
             self._updateStyles(maxHeight + 10);
@@ -1282,6 +1284,7 @@
             self._triggerChangeEvent(forceNotify);
 
             self.grid.endUpdate();
+            self.grid.cleanNodes();
 
             var nestedGrids = o.find('.grid-stack');
             if (nestedGrids.length && event.type == 'resizestop') {


andrei-gheorghiu:
I'd like to point out the above fix doesn't fix all cases of items disappearing, only the ones caused by moving alternatively items from different containers. It's still possible to delete components by dropping other components on top of them. While trying to fix this, I couldn't help but notice listeners stacking on items, which I believe to be the root of the problem. This lib would probably benefit a lot from an overhaul of the events binding and unbinding by either: a) always making sure a binding is not done twice, b) providing a method to clean and remap bindings on event (which is pretty much a special case of a). or c) bind on containers, only once rather than on items (also special case of a)).
Cheers.

radiolips:
@andrei-gheorghiu Thanks for that note. I believe you're correct about the duplication., and I agree that everything needs to be checked again with those events. We're working through removing jQuery, anyway.

radiolips:
Closing this as I believe changes in the last several months (including one from this morning) have resolved the referenced issue. Thank you for putting in this PR! Apologies for not getting to it sooner.

PR #945, opened 2018-10-08
Export on the correct scope for each environment.

### LINKED ISSUE #643: When used in browserify, gridstack is not correctly exported, which breaks the gridstack jquery drag plugin
I'm trying to move from 0.2.x to 0.3.0, however the module exporting code is not compatible with Browserify (or probably any CommonJS system).

The jQueryUI drag and drop plugin never correctly initializes.

This is because it is calling 
```javascript
try { GridStackUI = require('gridstack'); } catch(e) {}
```

Which always returns an empty object, because the `gridstack.js` file **never actually exports anything**, but instead "exports" to the global `window` scope, as seen in `gridstack.js` line 20:

```javascript
var scope = window;
```

And `gridstack.js` line 1729:

```javascript
scope.GridStackUI = GridStack;
```

`exports` is never actually used, so calling `require('gridstack')` will then always return an empty object.

So I think the module loading code for both `gridstack.js` and `gridstack.jQueryUI.js` should be altered to export using the `module.exports` CommonJS pattern.

I think a good way to do this is looking at [how backbonejs does this](https://github.com/jashkenas/backbone/blob/master/backbone.js#L8).

jaapz:
### Description

RequireJS and CommonJS now export on the `exports` module. `window`
exports in these environments are now removed no longer available.

Other environments still export on the `window`.

This fixes #643.

I think you should pay special attention to the fact that exporting to the `window` object is now gone on RequireJS and CommonJS exports, as they now correctly export just in their preferred `exports` module. This is probably a major breaking change, might be good to update the readme or whatever.

Do you need me to update a changelog too, before merging? I also ran `npm build` so the files in the `dist` directory are also updated.

### Checklist


jaapz:
It's reporting decreased coverage because I added a new argument to the module factory.

Issue #713, opened 2017-07-11
Custom Responsive Sort Order


gooeyideas:
I am having a hard time finding documentation for how to control the responsive grid.

How can I keep built-in responsiveness without the having the framework reorder the markup?

The markup structure that I am using gridstack with is already in the order that I want the stacked responsive grid to be in. The logic that gridstack does to sort the responsive tiles is undesirable.

radiolips:
@gooeyideas Are you talking about one column mode - when the window is small (ie on a mobile device), the widgets go into a single column. Is that what you're referring to?

gooeyideas:
Yes. One column mode removes the widgets and re-orders them when they are added back to the Dom to display for mobile. I would like this to be configurable. I am building an app that allows for a mobile configuration of a layout that is somewhat independent of the desktop layout.

radiolips:
I see. Well, if you're trying to accomplish that, you can handle that yourself. You can check to see if the grid is in one column mode, and you can store the information separately. Then, when you load (or modify) the grid, you could set the location using those values. Because gridstack is dynamic, it intentionally tries to exclude movement in one column mode.

gooeyideas:
You are misunderstanding the problem. The layouts can be configured easily. It is the display using gridstack that is problematic. I have written a fair amount of code to accommodate it but I believe that this extra code should not be necessary given what I am trying to accomplish.

This problem is that Gridstack assumes that the one-column order should always be determined by the widget's relative position in the grid. Additionally the logic that Gridstack uses to determine the order of widgets in One Column mode is thus collapse by "Left to Right" then "Top to Bottom" and as far as I can tell this is not configurable.

Here is an [example ](https://codepen.io/GooeyIdeas/pen/ZJNpWo?editors=1100) of the current behavior for refernce and clarification:
Given this HTML markup:
![image](https://user-images.githubusercontent.com/6644447/30118635-253664da-92e9-11e7-8d58-42c1cc948916.png)

Gridstack can render:
![image](https://user-images.githubusercontent.com/6644447/30119768-6918c000-92ec-11e7-9ac7-9d68e797f57e.png)

And in one-column mode Gridstack will render:
![gridstackmobile](https://user-images.githubusercontent.com/6644447/30119367-17311a4a-92eb-11e7-9050-8493ccf2a65c.png)

For my situation I would like a flag that disables Gridstack's 'One Column Sorting Logic' and allow them to display in the order from the original markup. 

Then my example could render in the proper order like this:
[...]

PR #966, opened 2019-01-25
Adding Support for oneColumnModeSort Options (Custom Responsive Sort Order)

gooeyideas:
…
### Description
These are changes implementing the requested changes from issue #713.

oneColumnModeSort option for sorting that also respects the existing `rtl` property.

`oneColumnModeSort: "row | row-reverse | column | column-reverse | none"`

- `none`: Sort by the order the widget's appear in the markup
- `row`:  (default) Sort first by **y** values (low to high) then sort widgets with the same y value by their **x** value (based on the `rtl` setting)
- `row-reverse`: Sort first by **y** values (high to low) then sort widgets with the same **y** value by their **x** value (based on the `rtl` setting)
- `col`: Sort first by **x** values (based on the `rtl` setting) then sort widgets with the same **x** value by their **y** value (low to high)
- `col-reverse`: Sort first by **x** values (based on the `rtl` setting) then sort widgets with the same **x** value by their **y** value (high to low)

Please explain the changes you made here. Include an example of what your changes fix or how to use the changes: **Please refer to #713 for more info**

### Checklist
- [X] All tests passing (`npm test`)
**I am not sure how to update the doc/README.md with the auto jsdoc generation tool. Is there anything else I need to add?**


src/gridstack.js
```diff
@@ -677,9 +681,15 @@
             cellHeightUnit: 'px',
             disableOneColumnMode: opts.disableOneColumnMode || false,
             oneColumnModeClass: opts.oneColumnModeClass || 'grid-stack-one-column-mode',
+            oneColumnModeSort: opts.oneColumnModeSort || 'row',
             ddPlugin: null
         });
 
+        this.opts.oneColumnModeSort = this.opts.oneColumnModeSort.toLowerCase();
+        if (!String('row | row-reverse | column | column-reverse | none').includes(this.opts.oneColumnModeSort)) {
+            this.opts.oneColumnModeSort = 'row';
```
adumesny: nice to keep original sort default. It's the most logical one really.

src/gridstack.js
```diff
@@ -777,19 +787,25 @@
                 }
                 self.container.addClass(self.opts.oneColumnModeClass);
                 oneColumnMode = true;
+                var oneColumnModeSort = self.opts.oneColumnModeSort;
+                if (oneColumnModeSort !== 'none') {
+                    var dim = oneColumnModeSort.includes('column') ? -1 : 1;
+                    var dir = dim > 0 && self.opts.rtl ? -1 : 1;
+                    dir = oneColumnModeSort.includes('reverse') ? dir * -1 : dir;
```
adumesny: code could use some comment explanation. rather hard to parse. for one thing you compare dim > 0, yet set value 1:-1 above so why not use those 2 values ?

adumesny:
"I am not sure how to update the doc/README.md with the auto jsdoc generation tool. Is there anything else I need to add?"
I believe README.md is manually updated - that's what I've been doing. Format is pretty self explanatory (and some tools like Visual Studio Code will render a pretty form of it so you can verify your tag formatting before checking it in).

gooeyideas:
Updated the original PR but I still need to go in and update the README, I will try to find some time to do that next week.

adumesny:
@gooeyideas any chance you might update this PR to the current code base ? we're actively accepting review and releasing frequent builds right now...

adumesny:
@gooeyideas I completely changed how we position widgets when changing column sizing (1 being the easy case really) with #1098. While this doesn't invalidate the need to have 'custom' order (dom vs row which is what I do hopefully corecly now) you may wan to check the latest behavior. And adjust your review if you still feel the need to have even more control...
having saved layouts could also mix with that you have (way for user to set those as opposed to gridstack generating them)

adumesny:
@gooeyideas please update the review. I don't think we should support all the modes, but rather having 1 column use left-to-right-top-bottom (default) or DOM order (so you can specify a very different 1 column layout (ideally with hidden items as well).

The other directions (RTL for example) should be tied to a generic gravity that affects all things (drag&drop, moving) rather than one off 1column thing IMO.

I would love to give you credit for the DOM order at least.

gooeyideas:
Sure thing, for missing the last few messages. I recently discovered that my company changed some of our email rules so that github alerts were being treated as spam. I will try get around to this before the end of this week. Sorry again.

adumesny:
great I look forward to it! are you ok doing just the DOM vs default row order we have now and we can add the others later or tie them to a generic gravity of the grid settings ?

I'm thinking that DOM order makes sense as way to specify a custom 1 column layout, whereas x,y are used for multi column layouts.

I've changed the code to support a robust multi column layout where we store each config (so going from 12 to 1 (or anything in between) and back would preserve the original layout, but have not worked on oneColumn yet, and was hoping to get your change in before I do.

adumesny:
@gooeyideas supporting just DOM order (vs default row sorting) is getting much trickier than I expected now that `oneColumnMode` is just `setColumn(1)` to allow full editing (resize, move, insert) and better heuristics as we don't re-order dom elements during editing so changes will not last.

I may try in sandbox but punt on this until we support per column layout instead in the file format (what I create in memory).

adumesny:
I didn't hear back so closing this as I ended up doing #1126 instead. Thanks for the initial work and let me know (in #713) if I missed anything

PR #974, opened 2019-02-11
Don't overwrite globals when in a modular environment.

garvank:
### Description
Requiring `jQuery` and `_` without a `var` declaration were overwriting the global implementations of those libraries if they exist.

Because `jQuery` and `_` can be customized and configured, it would be best to keep internal dependencies isolated.

I have a 2018 MBP and `npm test` does not run for me -- Karma simply times out. Since this falls back to the global versions, I don't foresee any tests failing. But please let me know if I'm missing a step to get tests running locally.

### Checklist

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -9,10 +9,13 @@
     if (typeof define === 'function' && define.amd) {
         define(['jquery', 'exports'], factory);
     } else if (typeof exports !== 'undefined') {
-        try { jQuery = require('jquery'); } catch (e) {}
-        factory(jQuery, exports);
+        var jQueryModule;
+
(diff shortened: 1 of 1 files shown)


garvank:
Hmm, happy to condense down to fewer lines to avoid a coverage decrease, but I think it's still worth it to not overwrite globals if we can.

radiolips:
Hey @garvank – We just removed lodash (thanks @adumesny !). Could you update this to only check for jQuery, then I'll merge? I agree that it should avoid overwriting global when possible.

garvank:
@radiolips thanks for circling back to this! Updated as requested :)

PR #978, opened 2019-03-07
Renamed resizestop event name on API docs

fedesnieg:
The event name was wrong on the API docs. The correct event name is resizestop


adumesny:
No it is not named wrong (maybe 'resizedone' would been clearer?) - see https://github.com/gridstack/gridstack.js/issues/398
gs has to have a custom event that is guaranteed to be AFTER the normal resizestop event sent in which we update our sizes, otherwise you might get called first. I'll update the doc.

radiolips:
Thanks for pointing this out, @fedesnieg , and thanks @adumesny for taking care of updating the docs.

Issue #1017, opened 2019-10-14
Widgets initiated with undefined data-gs-x and data-gs-y glitch


SomMeri:
If the widgets do not have initial data-gs-x and data-gs-y specified, widgets are rendered with wrong size. They are also all located on one place.

## Your environment
* version of gridstack.js: 0.4.0 <- it worked correctly in previous versions
* which browser and its version: all of them

## Steps to reproduce + Actual behaviour
Working demo: https://jsfiddle.net/2p7a9wjg/

Both widgets are created without data-gs-x and data-gs-y:
````html
<div class="grid-stack">
    <div class="grid-stack-item upper"
        data-gs-width="2" data-gs-height="2">
            <div class="grid-stack-item-content"></div>
    </div>
    <div class="grid-stack-item"
        data-gs-width="2" data-gs-height="2">
            <div class="grid-stack-item-content"></div>
    </div>
</div>
````

Only one widget is visible and it is small.

## Expected behaviour
This is how it should look like - rendered with 0.3.0: https://jsfiddle.net/m3fxs7u1/4/

Both widgets are visible and they are big.

adumesny:
actually if you label them 'item 1' and 'item 2' you will see in rev3
[2]
[1]

while in your fix you get
[1]
[2]

I think ideally not specify a position should land you
[1] [2]
instead of both trying to go space (0,0) as you have...

the overlap and wrong size is incorrect in v0.4 and v0.5 (get inf loop error) so we NEED a fix, don't think this is best fix.

radiolips:
@SomMeri Thanks for logging this and thanks for putting in a PR! @adumesny put in a slightly different version, which I think will help solve the problem. Try this out with our latest develop branch version! Please reopen or log a new ticket if your problem still isn't resolved.

PR #1018, opened 2019-10-14
Widgets initiated with undefined data-gs-x and data-gs-y glitch #1017

SomMeri:
### Description
This fixes the https://github.com/gridstack/gridstack.js/issues/1017 

The modified line originally ended up with `NaN` as a result. That caused the widget to not be initialized later on [this line](https://github.com/SomMeri/gridstack.js/blob/1b95810c4fefc319c3c5faf0b117f5d507ea42a9/src/gridstack.js#L396). This later caused the styles not to be initialized [in this loop](https://github.com/SomMeri/gridstack.js/blob/1b95810c4fefc319c3c5faf0b117f5d507ea42a9/src/gridstack.js#L1109)

Here is how compiled version that works: https://jsfiddle.net/30qfrzcd/

### Checklist
- [x] Extended the README / documentation, if necessary <- not necessary

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1388,8 +1388,8 @@
 
         el.addClass(this.opts.itemClass);
         var node = self.grid.addNode({
-            x: parseInt(el.attr('data-gs-x'), 10),
-            y: parseInt(el.attr('data-gs-y'), 10),
+            x: parseInt(el.attr('data-gs-x'), 10) || 0,
+            y: parseInt(el.attr('data-gs-y'), 10) || 0,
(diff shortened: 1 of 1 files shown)


adumesny:
thank you for posting this with running code, I will take a look at your review.

adumesny:
I believe #1030 is the correct way to fix this instead. Missing x,y => auto-placement, NOT (0,0) which is a valid position. having items stack up (differently from v0.3) isn't good either.

radiolips:
Closing because this should now be fixed with #1030 . Thanks for putting in the original fix, @SomMeri !

Issue #999, opened 2019-06-11
wrong counting of minHeight for resizable widget


miroslav-peterka:
At line

https://github.com/gridstack/gridstack.js/blob/17fcce8dc2808a941e683c9123e40664440bc819/src/gridstack.js#L1281

is some mistake, IMHO because you forgot to take in equation vertical margin.

example:
cellHeight = 80px and vertical margin = 20px
item height is 3, minHeight is 2

minHeight in pixels should be 80 x 2 + 20 = 180px. but at pointed line is counted number 188px

80*3 + 2*20 = 280
280 / 3 = 93.3 ~= 94
94 * 2 = 188px

radiolips:
@miroslav-peterka that may certainly be correct. Could you provide an example that I could take a look at?

adumesny:
@miroslav-peterka you are absolutely right - here is the before when you can't drag smaller and after my fix (coming up) to take margin into account.

before
![two_grids_before](https://user-images.githubusercontent.com/3496166/68261810-19b76d80-fff6-11e9-9a01-dd2783757226.png)

after
![two_grids_after](https://user-images.githubusercontent.com/3496166/68261812-1c19c780-fff6-11e9-8a5b-b498bf601b7a.png)

adumesny:
fixed in upcoming 0.5.2 release

PR #1024, opened 2019-11-06
onStartMoving() fix to minHeight

adumesny:
### Description
*method was not correctly counting margin when figuring out minHeight during a drag
(see #999)

Note: many other places don't handle margin, both height and width of cells.
Should probably be fix globally. WidthMargin is not
in opts though but CSS, making it harder.

### Checklist

I ran two.html sample app to make sure things scaled up/down correctly

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1278,8 +1278,9 @@
             self.grid.cleanNodes();
             self.grid.beginUpdate(node);
             cellWidth = self.cellWidth();
-            var strictCellHeight = Math.ceil(o.outerHeight() / o.attr('data-gs-height'));
-            cellHeight = self.container.height() / parseInt(self.container.attr('data-gs-current-height'));
+            height = o.attr('data-gs-height');
+            verticalMargin = self.opts.verticalMargin;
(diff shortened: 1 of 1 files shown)


src/gridstack.js
```diff
@@ -1278,8 +1278,9 @@
             self.grid.cleanNodes();
             self.grid.beginUpdate(node);
             cellWidth = self.cellWidth();
-            var strictCellHeight = Math.ceil(o.outerHeight() / o.attr('data-gs-height'));
-            cellHeight = self.container.height() / parseInt(self.container.attr('data-gs-current-height'));
+            height = o.attr('data-gs-height');
+            verticalMargin = self.opts.verticalMargin;
+            cellHeight = Math.ceil((o.outerHeight() - (height - 1) * verticalMargin) / height);
```
adumesny: if I make this `var cellHeight` (like strictCellHeight was) then we get an infinite loop stack error between moveNode() and _prepareNode() !
Not sure why... I don't think it's calling the method and we do things like this in other places:
var cellHeight = self.cellHeight();

adumesny:
will do a more global fix instead.

PR #1035, opened 2019-11-11
onStartMoving() fix to minHeight (#999)

adumesny:
### Description
*method was not correctly counting margin when figuring out minHeight during a drag
(see gridstack#999)
* fixed cellHeight() getter to correctly returns the value and checked all places that use to adjust for margins

### Checklist

I tested two.html as well as my app that doesn't a lot of drag&drop and made sure things resized correctly and no new issue, but min size is now correct (see #999 screen shots)


radiolips:
`opts.verticalMargin` can be a string, so those references will need to be changed to something that returns a number.

adumesny:
I changed one of the demo (two.html) to have `verticalMargin: "2rem"` in the opts and sure enough gridstack constructor calls `this.verticalMargin(this.opts.verticalMargin, true);` and the only other way to set is to also call `verticalMargin(val)` - in there we do extract he number vs the unit so opts.verticalMargin will NEVER store a string.

Now mind you we fail to convert the units to actual pixels so '2rem' = '2px' = '2em' which is clearly incorrect, but that's a separate issue (maybe the CSS is update for other values, but verticalMargin is pixel code and therefore wrong.

Issue #907, opened 2018-06-01
(Suggestion) addWidget to accept object instead of bunch of parameters


n1ghtie:
Hi, first of all THANK YOU for amazing plugin, loving it so far.
Working on a project that allows to create custom dashboards and need to create widgets dynamically (grid-stack-items).
I would be great that instead of putting each parameters separately, maybe to do an object so addWidget would look like this?

```javascript
var options = {
    x: 10,
    y: 10,
    width: 5,
    height: 5,
    ...
}
addWidget(el, options);
```
Why? Well lets say that if i want to have min-height, with current structure I need to supply min-width and max-width, and i don't really want that.

kylehovey:
In the mean time, here is a snippet that I use in my code:

```javascript
const genId = () => Math.random().toString(36).substr(2, 9);

const widget = stack => ({
  id = genId(),
  element,
  xPos = 0,
  yPos = 0,
  width = 1,
  height = 1,
  minWidth = 0,
  maxWidth = 10,
  minHeight = 0,
  maxHeight = 10,
  autoPos = false
} = {}) => stack.addWidget(
  element,
  xPos,
  yPos,
  width,
  height,
  autoPos,
  minWidth,
  maxWidth,
  minHeight,
  maxHeight,
  id
);
```

adumesny:
@n1ghtie I agree. We should add it soon.

adumesny:
in the upcoming 0.5.2 release

PR #1038, opened 2019-11-13
addWidget(el, options) with object param

adumesny:
### Description
fix for #907 
* new addWidget(el, options) to pass object so you don't have to spell 10 params
* updated TS file, test case, some samples
* also renamed IGridstackOptions to GridstackOptions (angular/TS style guide avoid cap I for interfaces, the others were correct already).

### Checklist

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -394,7 +394,7 @@
 
   GridStackEngine.prototype._prepareNode = function(node, resizing) {
     node = node || {};
-    // if we're missing position, have grid position us automatically (before we set them to 0,0)
+    // if we're missing position, have the grid position us automatically (before we set them to 0,0)
     if (node.x === undefined || node.y === undefined) {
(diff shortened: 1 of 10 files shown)


src/gridstack.js
```diff
@@ -1435,8 +1435,13 @@
     }
   };
 
-  GridStack.prototype.addWidget = function(el, x, y, width, height, autoPosition, minWidth, maxWidth,
-    minHeight, maxHeight, id) {
+  GridStack.prototype.addWidget = function(el, x, y, width, height, autoPosition, minWidth, maxWidth, minHeight, maxHeight, id) {
+
+    // instead of passing all the params, the user might pass an object with all fields instead, if so extract them and call us back
+    if (typeof x === 'object') {
```
creage: this will fail if we pass `null` as a value of `x`, as `typeof null === 'object'` returns true.

Issue #998, opened 2019-05-27
quick move the item between different grid-stack will assert error


liuyisnake:
## Subject of the issue
quick move the item between grid-stack will assert exception

## Your environment
* version of gridstack.js
1.0.0(develop) and 0.3.0(master)
* which browser and its version
chrome

## Steps to reproduce
use the demo of gridstack two.html could duplicate the issue
1.create tow grid-stack(grid-stack1 and grid-stack2) and several items
2.move one  item from grid-stack1 to grid-stack2
3.drag the item back to grid-stack1 and hover other draggable itmes, or quick move the item between two grid-stack. 

## Expected behaviour
work as normal

## Actual behaviour
it will assert error 

gridstack.js:660 Uncaught TypeError: Cannot read property 'attr' of null
    at gridstack.js:660
    at Mt (lodash.min.js:9)
    at Function.Te (lodash.min.js:38)
    at GridStackEngine.onchange (gridstack.js:653)
    at GridStackEngine._notify (gridstack.js:321)
    at GridStackEngine.addNode (gridstack.js:370)
    at HTMLDivElement.dragOrResize (gridstack.js:1088)
    at $.<computed>.<computed>._trigger (jquery-ui.js:799)
    at $.<computed>.<computed>._trigger (jquery-ui.js:6378)
    at $.<computed>.<computed>._trigger (jquery-ui.js:407)

i try to trace the dropout and dropover event of jquery-ui
it seems as below if drag the item from grid-stack2 back to grid-stack1 will trigger the dropout event but doesn't trigger the dropover event of grid-stack1 (and onDrag also will be triggered)

abrorabdukadirov:
hi liuyisnake did you solve this problem? I have the same problem

radiolips:
Thanks @liuyisnake . I'll look into this.

radiolips:
@liuyisnake @abrorabdukadirov I believe this is resolved on the latest `develop` branch. Please give it a try. Reopen the ticket if you still have an issue.

PR #1046, opened 2019-11-16
Bugfix/1043

### LINKED ISSUE #1043: Multiple Grids Problem on moving items
## Subject of the issue
Lets say we have two grids: grid A and grid B.
Each grid has some items. Lets say grid A contains item1 and grid B contains item2.
Whenever I move an item on the first one it moves as it is supposed to. 

We assume the following sequence of events: 
event 1: I move item 1
event 2: I move item 2
event 3: I move item 1
event 4: I move item 2
(Moves take place inside each grid and not drag-drop to one another)

While on event 3 (moving item 1 for second time), placeholder doesn't appear.
Also, on start of event 4 (moving item 2 for second time), item 1 gets disappeared. 

This can be checked on [Two grids demo](https://gridstackjs.com/demo/two.html)

Is this a known issue? May it be fixed?

*When acceptWidgets property is not defined, everything works smoothly. 

You have done a great job btw.

## Your environment
* gridstack 0.5.2 (also tested on 0.4 and the same thing happens)
* Firefox 70.0.1

radiolips:
### Description
Store information on node about whether widget is hovering over a different grid. Required to get around jQueryUI issue. Fixes #1043 

### Checklist
- [x ] Created tests which fail without the change (if possible)
- [x ] All tests passing (`yarn test`)
- [ x] Extended the README / documentation, if necessary


adumesny:
also fix #998 ? added for reference

PR #1053, opened 2019-11-18
grid options renamed to `columns` & `maxRows`

adumesny:
### Description
* grid options `width` is now `columns`, and `height` is now `maxRows` which match what they are.
* Old names are still supported for now (with console warnings)
* updated test (required), samples (optional), readme
* re-wrote entire custom column section. Hopefully much easier to find/use now.

Created a sample for #810 with 30 columns, but not seeing issues in that PR (issue rounding cellWidth() with large # of columns). I was able to move and resize and things lined up as expected even when total width / 30 was close to .5 off 
@nulen please update with info.

In the meantime I found those issues:
* fix README talking about custom columns #
* fix SASS code to handle >12 columns as you need to override stack-item min-width (we have 8.3333% otherwise which is 12 column!)
* while testing I incorrectly passed 'auto' for gridItem width (it's a cellHeight on grid options instead)
so fixed code to handle wrong strings/garbage passed for x,y,w,h as I was getting inf loop instead.
* updated test cases

### Checklist
also ran every demo samples, and the new 30 columns sample as well.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -104,10 +104,10 @@ gridstack.js API
 
 ## Item attributes
 
-- `data-gs-x`, `data-gs-y` - element position. Note: if one is missing this will `autoPosition` the item
-- `data-gs-width`, `data-gs-height` - element size
-- `data-gs-id`- good for quick identification (for example in change event)
-- `data-gs-max-width`, `data-gs-min-width`, `data-gs-max-height`, `data-gs-min-height` - element constraints
+- `data-gs-x`, `data-gs-y` - (number) element position in row/column. Note: if one is missing this will `autoPosition` the item
+- `data-gs-width`, `data-gs-height` - (number) element size in row/column
+- `data-gs-id`- (number | string) good for quick identification (for example in change event)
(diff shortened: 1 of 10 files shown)


doc/README.md
```diff
@@ -93,13 +94,12 @@ gridstack.js API
...
   * an integer (px)
   * a string (ex: '2em', '20px', '2rem')
-- `width` - amount of columns (default: `12`)
 
 ## Grid attributes
 
 - `data-gs-animate` - turns animation on
-- `data-gs-width` - amount of columns. Setting non-default value must be supported by equivalent change in CSS, [see docs here](https://github.com/gridstack/gridstack.js#change-grid-width).
-- `data-gs-height` - maximum rows amount. Default is `0` which means no maximum rows.
+- `data-gs-columns` - amount of columns. Setting non-default value must be supported by equivalent change in CSS, [see docs here](https://github.com/gridstack/gridstack.js#change-grid-columns).
```
adumesny: do we care about migrating those attr ? also not clear if that and options is set which ones wins.

PR #1067, opened 2019-11-22
Advanced demo page added

DigitalJo:
I''ve added a new page on the demo folder with the advanced example of the gridstack website

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ bower_components
 coverage
 dist
 node_modules
+/.vs
--- a/demo/advancedDemo.html
+++ b/demo/advancedDemo.html
@@ -1,0 +1,397 @@
+<html>
+<head>
+    <script src="https://code.jquery.com/jquery-3.4.1.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js"></script>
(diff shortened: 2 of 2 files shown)


adumesny:
@DigitalJo thanks for doing this, but it looks like you copied the entire runtime DOM instead of the source code - way to much HTML stuff. Also in the demo/ folder I've been using src relative files (since it helps us debug the code we write). Also data-gs-width="12" is old, and data-gs-column=12 is not needed (default).

If you want to check it in (and I think it's a good idea as it's different from two.html, I would switch to to the `hg-pages` branch and copy it out of index.html (the website landing page). You'll need to copy it somewhere before switching back to `develop` as they don't share any files.
I would rather have you get the credit for submit, else i can do it.

DigitalJo:
> u copied the entire runtime DOM instead of the source code - way to much HTML stuff. Also in the demo/ folder I've been using src relative files (since it helps us debug the code we write). Also data-gs-width="12" is old, and data-gs-column=12 is not needed (default).
> 
> If you want to check it in (and I think it's a good idea as it's different from two.html, I would switch to to the `hg-pages` branch and copy it out of index.html (the website landing page). You'll need to copy it somewhere before switching back to `develop` as they don't share any files.
> I would rather have you get the credit for submit, else i can do it.

thank you very much for your feedback.I will check it as soon as possible

adumesny:
I ended up checking one in for ya... thanks for the work!

PR #1098, opened 2019-12-19
new setColumn() heuristics

adumesny:
### Description
* complete re-write of the `setColumn()` code that moves/sizes widgets.
we now do a reasonable job positioning widgets (especially 1 column), and
we remember positions for each column layout to going from 12 -> 1 -> 12
will bring it exactly back for example!
* new column.html test case (now that it works) to be published.
* fixed bunch of bugs I found with overlapping widgets (going to 2 or 1 column)
* fixed bug where default (empty) option when adding a widget would render it at (0,0)
overlapping any existing widgets. added `readAttr()/_writeAttr()` to help there.
* unit test greatly updated with exact location expected of widgets (not that I agree with location)

better fix than #809

fixing responsive.html will come next...

### Checklist


adumesny:
thank you for the suggestion - got me intrigued in fixing it right (1 column which is trivial always bugged me being wrong). fixed by #1098

PR #1149, opened 2020-02-16
Cannot set property 'isOutOfGrid' of undefined

ronvanpol:
Adding widget to nested grid: Cannot set property 'isOutOfGrid' of undefined


adumesny:
@ronvanpol thank you for the code - I ended up debugging it deeper and found the actual fix - see #1150

closing this now.

PR #1157, opened 2020-02-23
Typescript + remove jQuery from API **BREAKING**

adumesny:
EARLY as I have NOT compiled/tested the TS files yet (need to change build system) but didn't want to loose 4+ days of work!
---------
BIG, BIG change
* converted all code to Typescript (still  some TODO)
* all calls/return take HTLMElement | string, not JQuery anymore
* the `change|removed|added|enable|disable` events are now using standard calls with CustomEvent
(rest are still JQ based) but all share basic same callback signature still.
* added grid.on(eventName, callback) / grid.off(eventName) to hide native JQ events mix
* updated all demos+doc to only plain DOM manipulation (only custom drag&drop jq UI remain)
* created how to upgrade to 1.0.0 section
* part of gridstack#1084

Also (API break)
* `setColumn(N)` is now `column(N)` and new `getColumn()`
* following getter had to change for TS `getCellHeight()`, `getVerticalMargin()`, `getFloat()`
* new `getRow()`

TODO:
* not compiling TS files yet (have to change build system) so work in progress..
* had to modify jquery and jquery-ui loading to work with webpack app, but only
gridstack.min.js gives me new error (.js is ok) so try to convert to TS and see...

RANT: I hate git... must have wasted half day doing a rebase from develop

### Description

### Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 dist/*.js
 demo/*
 spec/*
+src/jquery.js
 src/jquery-ui.js
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,13 @@
+{
+  "cSpell.words": [
+    "Pavel",
+    "Reznikov",
+    "dragdrop",
+    "dragstop",
+    "dropover",
+    "droppable",
+    "gsresize",
+    "gsresizestop",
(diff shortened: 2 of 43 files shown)


adumesny:
will redo this as Typescript only when ready...

PR #1169, opened 2020-02-27
fix resize not taking margin height

### LINKED ISSUE #1166: Resizing does not work as expected
## Subject of the issue
Resizing does not work as expected

## Your environment
1.0.0 and Chrome

## Steps to reproduce
https://pjc0247.github.io/bugtest/index.html

## Expected behaviour
Size should follow the user input.

## Actual behaviour
The object grows larger than expected.

gridstack

awjae:
### Description
fix #1166 issue

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -1296,7 +1296,7 @@
       // TODO: cellHeight = cellHeight() causes issue (i.e. remove strictCellHeight above) otherwise
       // when sizing up we jump almost right away to next size instead of half way there. Not sure
       // why as we don't use ceil() in many places but round() instead.
-      cellHeight = self.$el.height() / parseInt(self.$el.attr('data-gs-current-row'));
+      cellHeight = (self.$el.height() + grid.verticalMargin()) / parseInt(self.$el.attr('data-gs-current-row'));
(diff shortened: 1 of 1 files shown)


src/gridstack.js
```diff
@@ -1296,7 +1296,7 @@
       // TODO: cellHeight = cellHeight() causes issue (i.e. remove strictCellHeight above) otherwise
       // when sizing up we jump almost right away to next size instead of half way there. Not sure
       // why as we don't use ceil() in many places but round() instead.
-      cellHeight = self.$el.height() / parseInt(self.$el.attr('data-gs-current-row'));
+      cellHeight = (self.$el.height() + grid.verticalMargin()) / parseInt(self.$el.attr('data-gs-current-row'));
```
adumesny: should be self.verticalMargin() but I can change it later.

adumesny: while this fixes the issue, one downside is that now you need to stretch WAY past 50% point to get next size up. So that needs to be looked at as well.

adumesny:
@awjae thanks for the fix. I did #1170 to further improve on it.

PR #1173, opened 2020-02-27
Added minRow and row

RadoiAndrei:
### Description
Added `minRow` and `row` properties.
- `minRow`: Minimum amount of rows to generate, default is 0.
- `row`: Shortcut to have same value for `minRow` and `maxRow`.

### Checklist

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -75,6 +75,7 @@ gridstack.js API
   * a string (ex: '100px', '10em', '10rem', '10%')
   * 0 or null, in which case the library will not generate styles for rows. Everything must be defined in CSS files.
   * `'auto'` - height will be calculated to match cell width (initial square grid).
+- `row` - number of rows. This is a shortcut of writting `minRow: sameValue, maxRow: sameValue`.
 - `column` - number of columns (default: `12`) which can change on the fly with `column(N)` as well. See [example](http://gridstackjs.com/demo/column.html)
 - `ddPlugin` - class that implement drag'n'drop functionallity for gridstack. If `false` grid will be static. (default: `null` - first available plugin will be used)
 - `disableDrag` - disallows dragging of widgets (default: `false`).
@@ -86,6 +87,7 @@ gridstack.js API
 - `handle` - draggable handle selector (default: `'.grid-stack-item-content'`)
 - `handleClass` - draggable handle class (e.g. `'grid-stack-item-content'`). If set `handle` is ignored (default: `null`)
 - `itemClass` - widget class (default: `'grid-stack-item'`)
+- `minRow` - minimum rows amount. Default is `0`
 - `maxRow` - maximum rows amount. Default is `0` which means no maximum rows
 - `minWidth` - minimal width. If window width is less than or equal to, grid will be shown in one-column mode (default: `768`)
 - `oneColumnModeDomSort` - set to `true` if you want oneColumnMode to use the DOM order and ignore x,y from normal multi column layouts during sorting. This enables you to have custom 1 column layout that differ from the rest. (default?: `false`)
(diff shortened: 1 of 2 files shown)


src/gridstack.js
```diff
@@ -296,10 +296,12 @@
 
   var idSeq = 0;
 
-  var GridStackEngine = function(column, onchange, float, maxRow, items) {
+  var GridStackEngine = function(row, column, onchange, float, minRow, maxRow, items) {
```
adumesny: I do not want to have a number of rows, unlike columns which are fixed and special % CSS layout. Min/MaxRow is more than enough to get everything you need. Also maxRow is not that well supported so I don't want to expose 'row' which is a superset. Please take out.

Also adding new param to new GridStackEngine() means all the test cases that use need updating - please fix them (under spec/) and make sure `npm test` still works after your changes AND the demo/ apps run. I know it's a lot but something I will have to do otherwise and I don't have extra cycle for this feature right now.

src/gridstack.js
```diff
@@ -1155,12 +1162,12 @@
 
   GridStack.prototype._updateContainerHeight = function() {
     if (this.engine._batchMode) { return; }
-    var row = this.engine.getRow();
+    var row = this.opts.minRow > this.engine.getRow() ? this.opts.minRow : this.engine.getRow();
```
adumesny: don't call methods twice! inefficient. either put the minRow in getRow() (see where it's used or here)

src/gridstack.js
RadoiAndrei: Im actually happy that it actually works. Will refactor tomorrow, I also need to get more familiar with the code. Also, is there any reasons why you use `var` instead of `let` and `const`?

adumesny: didn't write that legacy code! I'm actually converting to typescript (as I willl have hand pick all those changes since 1.0.0) as I hate JS spaghetti code.

RadoiAndrei:
I implemented minRow how @adumesny told me to do it. Now if `row` is set, it will replace both the values of `minRow` and `maxRow` in opts.

src/gridstack.js
```diff
@@ -1156,6 +1164,9 @@
   GridStack.prototype._updateContainerHeight = function() {
     if (this.engine._batchMode) { return; }
     var row = this.engine.getRow();
+    if (row < this.opts.minRow) {
```
adumesny: yep, that is the right place to put it. as you can see you can also accomplish this with CSS `min-height` on the grid div, but setting minRow can be easier for some.

src/gridstack.js
```diff
@@ -709,8 +715,10 @@
     var isNested = this.$el.closest('.' + opts.itemClass).length > 0;
 
     this.opts = Utils.defaults(opts, {
+      row: parseInt(this.$el.attr('data-gs-row')) || 0,
       column: parseInt(this.$el.attr('data-gs-column')) || 12,
-      maxRow: parseInt(this.$el.attr('data-gs-max-row')) || 0,
+      minRow: parseInt(this.$el.attr('data-gs-row')) ? parseInt(this.$el.attr('data-gs-row')) : parseInt(this.$el.attr('data-gs-min-row')) || 0,
```
adumesny: please extract this.$el.attr('data-gs-row') into a var right above - easier to read/parse. maybe `minRowAttr`

src/gridstack.js
```diff
@@ -693,6 +693,12 @@
 
     opts = opts || {};
 
+    // if row property exists, replace minRow and maxRow
```
adumesny: move that down - feel weird to be first thing we check... like right after 715 where we set opts values.

doc/README.md
```diff
@@ -75,6 +75,7 @@ gridstack.js API
   * a string (ex: '100px', '10em', '10rem', '10%')
   * 0 or null, in which case the library will not generate styles for rows. Everything must be defined in CSS files.
   * `'auto'` - height will be calculated to match cell width (initial square grid).
+- `row` - number of rows. This is a shortcut of writting `minRow: sameValue, maxRow: sameValue`.
```
adumesny: great, but all those need to be alphabetized please. also `row` should say something like:

`row` - fixed number of rows which prevents the grid from changing as items are added/removed/moved, Shortcut for setting `minRow: N, maxRow: N`. default is `0` no constrain.

adumesny:
thanks, the last code version looks a lot better. I'll accept it and tweak the doc.

PR #1187, opened 2020-03-05
Add polyfill for CustomEvent to fix #1186

### LINKED ISSUE #1186: CustomEvent IE throws "Object doesn't support this action"
## Subject of the issue
Object doesn't support this action error thrown

## Your environment
* 1.1.0 version of gridstack.js
* IE 11 (assuming IE 9+)

## Steps to reproduce
IE not supported by jsfiddle

## Expected behaviour
No error

## Actual behaviour
Error

phil-blais:
### Description

### Checklist

### DIFF
--- a/src/gridstack-poly.js
+++ b/src/gridstack-poly.js
@@ -5,6 +5,18 @@
  * gridstack.js may be freely distributed under the MIT license.
 */
 
+// https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent
+if (typeof window.CustomEvent !== "function") {
+  function CustomEvent (event, params) {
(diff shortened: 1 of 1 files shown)


src/gridstack-poly.js
```diff
@@ -5,6 +5,18 @@
  * gridstack.js may be freely distributed under the MIT license.
 */
 
+// https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent
```
adumesny: site has it wrapped in a function it calls.. maybe to scope the CustomEvent var ? I would do the same please.

phil-blais: My mistake. Thought the whole poly file was scoped for reason. Updated.

PR #1202, opened 2020-03-10
TS: Replace some `jQuery`

btecu:
Work in progress. Replaced some `jQuery` with native while keeping the `jQuery` calls around for reference.


adumesny:
thank you! I will continue on this...

PR #1205, opened 2020-03-11
TS: Fix lint command on Windows

btecu:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,7 +16,6 @@ module.exports = {
   rules: {
     'indent': ['error', 2],
     'max-len': ['error', 180],
-    'no-trailing-spaces': 'error',
-    'prefer-const': 0
+    'no-trailing-spaces': 'error'
   }
 };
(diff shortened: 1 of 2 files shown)


adumesny:
sweet, thanks!

.eslintrc.js
```diff
@@ -16,7 +16,6 @@ module.exports = {
   rules: {
     'indent': ['error', 2],
     'max-len': ['error', 180],
-    'no-trailing-spaces': 'error',
-    'prefer-const': 0
```
adumesny: should that be true instead ? vscode was hitting at using const instead of let anyway... we can tighten the lint rules once we get it mostly working. forcing not `any` type would like force us to think hard about what types we passing around and fix jq vs element issues for one thing...

btecu: I prefer using `const` only for outside of class declaration, similar to `const` in other programming languages.

> For most variables, use let. For constants, use const. Constants should be declared at the top of modules and only in module scope.
My views of `const` vs `let` are aligned with this article: https://madhatted.com/2016/1/25/let-it-be

btecu: By default I believe `prefer-const` is set to `true`.
I turned removed the off flag because I saw you were changing `let` to `const`.

PR #1214, opened 2020-03-15
No more `$(el).children()`

btecu:
Small change to remove more jQuery dependencies, although I haven't tested it.

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -497,10 +497,12 @@ export class GridStack {
     let domNodes: GridStackNode[];
     if (column === 1 && this.opts.oneColumnModeDomSort) {
       domNodes = [];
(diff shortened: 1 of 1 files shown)


src/gridstack.ts
```diff
@@ -497,10 +497,12 @@ export class GridStack {
     let domNodes: GridStackNode[];
     if (column === 1 && this.opts.oneColumnModeDomSort) {
       domNodes = [];
-      this.$el.children('.' + this.opts.itemClass).each((index, el: GridItemHTMLElement) => {
-        let node = el.gridstackNode;
-        if (node) { domNodes.push(node); }
-      });
+      Array.from(this.el.children)
```
adumesny: thank you! I think there is a 4th one (checks for ':not(.' + this.opts.placeholderClass + ') which we can always do, and should be extracted into a convenience routine.
I can do that.

PR #1215, opened 2020-03-15
TS: added `getGridItems()`

adumesny:
continue @btecu work to extract `$(el).children()` out

### Checklist


mikehaas763:
@adumesny I have experience converting JS codebases to TS. Are you looking for any help?

adumesny:
@mikehaas763 absolutely! the `typescript` branch has all the code converted into Ts classes and most methods should now have arg signatures and many of the code re-written to TS **and** take jquery away from main code but there is still more to do...

For one thing the karma tests are not running anymore (I changed .js entry to use .ts files for code coverage as well) and wondering if that something you might  be more familiar with. Also I'm using webpack to generate a single `gridstack.all.js` and all the demos are finally working, BUT I'm also wondering if I should generate ES6 files or other format for other TS app to compile against (like my Angular8 based app I'm working on).

so yeah any help would be greatly appreciated!

PR #1241, opened 2020-03-23
Export version without jQuery

btecu:


btecu:
@adumesny I've rebased this and thanks to your latest work it's now green, the typescript are gone.

Also looks like the export is working:
```
                 Asset       Size               Chunks             Chunk Names
      gridstack-all.js    236 KiB        gridstack-all  [emitted]  gridstack-all

gridstack-no-jquery.js    115 KiB  gridstack-no-jquery  [emitted]  gridstack-no-jquery

```

Can you take a look and let me know if you'd like different exports or if this is enough?

adumesny:
thanks for doing this, but I'm not sure of the need for that other file right now as you will not get any drag&drop - not really useful (apart for static grids) until we have an HTML5 drag&drop replacement - which won't be until 3.x

PR #1244, opened 2020-03-29
Do not rely on options Object.prototype

### LINKED ISSUE #1195: Resizable options no longer working with ember hash helper Object.create(null)
After upgrading from 0.6.0 to 1.x, looks like setting the resizable options is no longer an option.
Using `yahoo/ember-gridstack`, although I suspect the issue is not with the addon but GridStack itself.

Here is how being invoked:
```
<GridStack @options={{hash resizable=(hash handles="e")}}>
  // items here
</GridStack>
```

Here is GridStack being initialized in the addon:
image

The error is not very useful:
image

If I remove `resizable` it works fine, except I can't change the `handles` value.

btecu:
Fixes #1195.

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -168,7 +168,7 @@
 
       sources.forEach(function(source) {
         for (var prop in source) {
-          if (source.hasOwnProperty(prop) && (!target.hasOwnProperty(prop) || target[prop] === undefined)) {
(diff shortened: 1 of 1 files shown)


adumesny:
@btecu can you redo this as it should go into 'develop' not 'master' (yeah I wish I could combine into just one as it's a pain to match them...

btecu:
@adumesny updated.

PR #1251, opened 2020-04-10
prevent removing another grid's item

### LINKED ISSUE #1250: Multiple Grids - Draggable destroy being called on non-existent element
## Your environment
gridstack.js 1.1.1
Chrome version 80.0.3987.163 (Official Build) (64-bit)

## Steps to reproduce
Create two grids, add stuff to first grid then add stuff to second grid, remove one by one each widget from the second grid then call grid.removeAll() on that second grid, you will get this error as the nodes for that grid is not empty yet the grid it's self has no widgets.

## Expected behaviour
What should happen is a check to see if the grid being cleared has remaining elements inside of it and whether or not the engine nodes list matches up or not, if not then empty the engine.nodes array so to match the actual status of the grid element.

## Actual behaviour
The second grid is empty after calling grid.removeWidget on each widget but the engine.nodes of the grid in question still contains a node which causes the error to fire.

My solution is to check if the grid is empty using jquery, if it is empty and the grid.engine.nodes isn't empty I empty it using array splice which resolves the issue.
`if(grids[1].engine.nodes.length>0 && $(bgGrid.el).is(':empty')) grids[1].engine.nodes.splice(0, grids[1].engine.nodes.length);`

This solution is good as even if this is fixed in the future it won't break peoples projects.
It may be important to note in my testing that the main grid (grids[0]) doesn't have this issue, only additional grids which I find a bit odd.

adumesny:
* fix #1250

### Checklist

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -36,6 +36,7 @@ Change log
 ## 1.1.1-dev (upcoming)
 
 - fix [1229](https://github.com/gridstack/gridstack.js/issues/1229) `staticGrid` no longer disable oneColumnMode
+- fix [1250](https://github.com/gridstack/gridstack.js/issues/1250) don't remove item from another grid
 - add `getGridItems()` to return list of HTML grid items
 
 ## 1.1.1 (2020-03-17)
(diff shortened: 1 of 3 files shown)


adumesny:
fixed in next release. thanks.

PR #1313, opened 2020-07-30
adding styleInHead option

gregid:
### Description
This change allows for keeping older behavior of gridstack where `<STYLE>` was added to `<HEAD>` instead of parentNode. The default behavior remains to be to use parentNode but in situations where `head` is preferrable just pass `styleInHead = true` to gridstack options.

### Checklist
- [X] Newly added tests are passing. There is one test that is currently failing on `gridstack.js/develop` though

### DIFF
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -733,6 +733,7 @@
       placeholderText: '',
       handle: '.grid-stack-item-content',
       handleClass: null,
+      styleInHead: false,
       cellHeight: 60,
       verticalMargin: 20,
       auto: true,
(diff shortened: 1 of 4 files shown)


src/gridstack.js
```diff
@@ -1092,8 +1093,9 @@
       Utils.removeStylesheet(this._stylesId);
     }
     this._stylesId = 'gridstack-style-' + (Math.random() * 100000).toFixed();
-    // insert style to parent (instead of 'head') to support WebComponent
-    this._styles = Utils.createStylesheet(this._stylesId, this.el.parentNode);
+    var styleLocation = this.opts.styleInHead ? false : this.el.parentNode
```
adumesny, line 1096: I would use null or undefined instead since parent is an element pointer (issue in typescript version) but works here in JS.

gregid: changed to `undefined`

adumesny:
thank you for contributing -  great job at documenting and testing this. I can make a new release soon if nothing else comes up..
I'll push your changes to 2.x typescript branch as well.

PR #1327, opened 2020-08-12
Update README.md

edudepetris:
### Description
Update the readme.md file  in order to map `x` with column and `y` with row

### Checklist


adumesny:
thanks if you could change the develop branch as well as that is the main one really... master is rather stale.

PR #1339, opened 2020-08-24
RTL Layout Support

sinap-se:
### Description
Added support for RTL layouts.

image

### Checklist

### DIFF
--- a/demo/right-to-left(rtl).html
+++ b/demo/right-to-left(rtl).html
@@ -58,7 +58,7 @@
       template:
         [
           '<div class="grid-stack" data-bind="foreach: {data: widgets, afterRender: afterAddWidget}">',
-          '   <div class="grid-stack-item" data-bind="attr: {\'data-gs-x\': $data.x, \'data-gs-y\': $data.y, \'data-gs-width\': $data.width, \'data-gs-height\': $data.height, \'data-gs-auto-position\': $data.auto_position}">',
+          '   <div class="grid-stack-item" data-bind="attr: {\'data-gs-width\': $data.width, \'data-gs-height\': $data.height, \'data-gs-auto-position\': $data.auto_position}">',
           '     <div class="grid-stack-item-content"><center><button data-bind="click: $root.deleteWidget">Delete me</button><br><h5 data-bind="text: index" /></center><br><p>lorem ipsum</p></div>',
           '   </div>',
           '</div> '
@@ -89,10 +89,10 @@
     };
 
     var widgets = [
-      {x: 0, y: 0, width: 2, height: 2, index: 1},
-      {x: 2, y: 0, width: 4, height: 2, index: 2},
-      {x: 6, y: 0, width: 2, height: 4, index: 3},
-      {x: 1, y: 2, width: 4, height: 2, index: 4}
+      {width: 2, height: 2, index: 1},
+      {width: 4, height: 2, index: 2},
+      {width: 2, height: 4, index: 3},
+      {width: 4, height: 2, index: 4}
     ];
 
     var controller = new Controller(widgets);
--- a/src/gridstack.d.ts
+++ b/src/gridstack.d.ts
@@ -394,17 +394,15 @@ interface GridStack {
 }
 
 /**
- * Defines the GridStack engine that does most no DOM grid manipulation.
+ * Defines the GridStack engine that does most DOM grid manipulation.
  * See GridStack methods and vars for descriptions.
  * 
  * NOTE: values should not be modified - call the GridStack API instead
  */
 interface GridStackEngine {
-  column: number;
-  float: boolean;
-  maxRow: number;
+  options?: GridstackOptions;
+  onChange(nodes: GridStackNode[], detachedNode: boolean): void;
   nodes: GridStackNode[];
-  getRow(): number;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


src/gridstack.js
```diff
@@ -296,7 +296,8 @@
 
   var idSeq = 0;
 
-  var GridStackEngine = function(column, onchange, float, maxRow, items) {
+  var GridStackEngine = function(opts, column, onchange, float, maxRow, items) {
```
adumesny, line 299: so column, float and maxRow are all grid options so it doesn't make sense to pass generic options there (for rtl) but others spelled out, You are breaking API anyway so might as well clean it up.

src/gridstack.js
```diff
@@ -515,17 +516,18 @@
   };
 
   GridStackEngine.prototype.addNode = function(node, triggerAddEvent) {
+    rtl = this.opts.rtl || false;
```
adumesny, line 519: var rtl

sinap-se:
> thank you so much for doing this!
> One other thing you forgot is updating the gridstack.d.ts definition file for GridStackEngine.
> 
> Also did you test drag&drop support to make sure widgets are added when dragged ? and what happens when you serialize the positions, do they start 0 on right side ? I'm guessing there is much more than this... also the resize handle need to be reversed. OneColumnMode support, changing column number... I don't believe this fix is enough, but a start.
> 
> you can also roll back your changes into 1 commit that makes it easier to see -
> git rebase -i HEAD~2 = edit the last 2 commits (change text, merge commits, etc...) - follow directions.
> git push -f
> 
> Also I'm doing all development in the 'typescript branch' and wasn't planning on releasing a 1.x anymore, so maybe I can convince you to do that in there instead and fix GridStackEngine constructor to take a GridstackOptions + callback instead of the spelled out params ?
> If not I can migrate them later...

Positions are based on the 0,0 being in the top left as usual so the 'first item' with a width of 2 would appear at x = 10, y = 0. Drag and drop works fine from what I can tell. Adding widgets was fine. These can both be verified in the RTL demo page.

I don't need the resize handle at the moment so I am not changing it yet. This incremental change is harmless, in my opinion, which is why I threw it back in to the pile.

I don't see why one column mode wouldn't work. I change the column count in my usage of the library so I am fairly confident it works fine. 

I can not accept typescript in to my project yet so I will not be doing this. I do have a plan to migrate at some point in the near future. In the short term, if I can not get this change in to the core 1.2.x product, I will likely have to fork and maintain my own.

src/gridstack.js
sinap-se: Resolved.

src/gridstack.js
sinap-se: Options object used for most params now. Signature is now GridStackEngine(opts, onChange, nodes)

sinap-se:
One column mode actually does not seem fine. Looking in to it.

sinap-se:
Single column mode does not work as I would expect. Looking in to it.

src/gridstack.d.ts
```diff
@@ -394,17 +394,15 @@ interface GridStack {
 }
 
 /**
- * Defines the GridStack engine that does most no DOM grid manipulation.
+ * Defines the GridStack engine that does most DOM grid manipulation.
```
adumesny, line 397: should be "most none DOM grid"

src/gridstack.d.ts
```diff
@@ -394,17 +394,15 @@ interface GridStack {
...
  * NOTE: values should not be modified - call the GridStack API instead
  */
 interface GridStackEngine {
-  column: number;
-  float: boolean;
-  maxRow: number;
+  options?: GridstackOptions;
+  onChange(nodes: GridStackNode[], detachedNode: boolean): void;
   nodes: GridStackNode[];
-  getRow(): number;
```
adumesny, line 407: getRow() is still a valid method.

demo/right-to-left(rtl).html
```diff
@@ -89,10 +89,10 @@ <h1>RTL Demo</h1>
     };
 
     var widgets = [
-      {x: 0, y: 0, width: 2, height: 2, index: 1},
-      {x: 2, y: 0, width: 4, height: 2, index: 2},
-      {x: 6, y: 0, width: 2, height: 4, index: 3},
-      {x: 1, y: 2, width: 4, height: 2, index: 4}
+      {width: 2, height: 2, index: 1},
```
adumesny, line 92: Did you have to remove x,y (same as auto: true). that's the thing adding rtl=true is NOT enough to switch between modes as all CSS and internal code is still (0,0) at top-left and not right side.

while you patch addNode() to flip X around, a lot more code (serialize, and switching column count live) will sort the nodes left->right and so that sortNode() and likely other parts will need to change. OneColumn also uses item sorting to figure out who's first (today it will be top-left).

sinap-se: Yeah, as I fiddle with the options there's a bunch of cases I did not consider. The interception logic is not aware of the change in direction, for example.

adumesny:
looking at the code (default to `auto`) looks like you need to set the style on the grid to direction = rtl.
```js
    if (this.opts.rtl === 'auto') {
      this.opts.rtl = el.style.direction === 'rtl';
    }
    if (this.opts.rtl) {
      this.el.classList.add('grid-stack-rtl');
    }
```
but no-where in the code do I see it being used otherwise... so not sure if there is more than CSS to make it work (ideally the CSS positions would be flipped) as I have not researched this.

PR #1341, opened 2020-09-01
Enable the UMD behavior for bundlers compatibility

### LINKED ISSUE #1243: gridstack.all 1.x import  failure
vue-element-admin  import  gridstack.all  failure

gridstack.js 1.1.1

in page.vue

import GridStack from '../../plug-in/gridstack/dist/gridstack.all.js'
options    =  {float: false};
this.grid  	=   GridStack.init(options);

show Error in mounted hook: "TypeError: _gridstackAll.default.init is not a function"

How can I use it correctly？ thanks

juchi:
### Description
Since v1.0.0 the UMD code at the top of the files has been disabled, preventing someone using a bundler to provide jQuery via the module system. Therefore if one's not including the gridstack.all.js directly in HTML, it is mandatory to have jQuery accessible globally via the `window` object.

I propose to enable back the UMD code in order to be able to use a bundler like webpack with gridstack v1

I have also modified the Readme file in order to reflect this change : gridstack.all.js should not be used with a bundler, instead we just need the gridstack lib files. A reference to the Gridstack object is retrieved via the import of gristack.js.

Will probably fix #1243 (I couldn't test this with a vue.js stack)

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,12 +71,34 @@ npm install --save gridstack
 
 ## Include
 
-after you install:
+* Using webpack, you can now import the files:
 
 ```js
-import 'gridstack/dist/gridstack.all.js';
+import 'jquery';
+import 'jquery-ui';
+
+import 'gridstack/dist/gridstack-poly'; // optional IE support
+import GridStack from 'gridstack/dist/gridstack';
+import 'gridstack/dist/gridstack.jQueryUI';
 import 'gridstack/dist/gridstack.css';
 ```
+
+If your project doesn't make use of jQuery or jQueryUI and you don't want to include these libs,
+you can use the lightweight files provided by Gridstack instead of the official ones.
+
+For this you have to define an alias in your webpack configuration:
+```
+module.exports = {
+  ...
+  resolve: {
+    alias: {
+      'jquery': 'gridstack/dist/jquery.js',
+      'jquery-ui': 'gridstack/dist/jquery-ui.js',
+    }
+  }
+}
+```
+
 * alternatively in html
 
 ```html
--- a/src/gridstack.jQueryUI.js
+++ b/src/gridstack.jQueryUI.js
@@ -5,14 +5,13 @@
  * gridstack.js may be freely distributed under the MIT license.
 */
 (function(factory) {
-  /* we compile this in so no need for required loading
   if (typeof define === 'function' && define.amd) {
     define(['jquery', 'gridstack', 'exports'], factory);
   } else if (typeof exports !== 'undefined') {
     try { jQuery = require('jquery'); } catch (e) {}
     try { gridstack = require('gridstack'); } catch (e) {}
     factory(jQuery, gridstack.GridStack, exports);
-  } else */{
+  } else {
     factory(jQuery, GridStack, window);
   }
 })(function($, GridStack, scope) {
--- a/src/gridstack.js
+++ b/src/gridstack.js
@@ -6,7 +6,6 @@
  * @preserve
 */
 (function(factory) {
-  /* [alain] we compile jquery with our code, so no need to 'load' externally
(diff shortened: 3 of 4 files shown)


adumesny:
@juchi thanks, I will take a look (I wasn't planning on releasing 1.x as I'm getting 2.0 out but might). 

Have you tried using gridstack versions of jquery and jquery-ui (which is 1/5th the size) and does it pick it up in your bundler, or do you need to provide your own version of either ones ?

Does all.js still compile and works for the HMTL demos ?

i recall having some issues with the UMD and webpack but haven't tried it again.

juchi:
The gridstack.all.js does compile and works with the HTML demos provided. The demos also still work with the files imported separately from "src" directory.

Thank for pointing out that the versions of jquery and jquery-ui provided are not the official ones, I didn't notice that.
I had made test with official jquery/jquery-ui libs so I have tested the stack again with your custom versions. It works fine as long as the user defines the "jquery" alias in his webpack config, ie:

```
  resolve: {
    alias: {
      jquery: 'gridstack/dist/jquery.js',
    }
  }
```
This is not needed when using the standard jquery package (as it defines the "jquery" module itself).

The same goes for jquery-ui, it needs to be imported in the project, by it can be the official one or the gridstack one. Both worked on my machine.

I have updated the readme file to reflect these details.

README.md
```diff
@@ -71,12 +71,34 @@ npm install --save gridstack
...
+import 'jquery';
+import 'jquery-ui';
+
+import 'gridstack/dist/gridstack-poly'; // optional IE support
+import GridStack from 'gridstack/dist/gridstack';
+import 'gridstack/dist/gridstack.jQueryUI';
 import 'gridstack/dist/gridstack.css';
 ```
+
+If your project doesn't make use of jQuery or jQueryUI and you don't want to include these libs,
```
adumesny, line 86: actually they can still use jquery in their code, it's the official latest one I include. Only thing is jquery-ui is a trimmed down version (1/5th the size) that has just what you need for drag&drop, so if you don't need all jquery-ui yourself you might want to include this smaller one. You can bring whichever jquery (1.8.3+ I believe) you want.

adumesny:
great, thank you! I will test this over the wkd and make a release. 

You might want to try the 2.0.0-rc2 (will do rc3 over the wkd with some packaging fixes) as well.

README.md
juchi: ok so it may be better for webpack to recommend jquery 1.8.3+ managed through npm and just mention the smaller jQueryUI, is that ok for you ?

juchi:
> You might want to try the 2.0.0-rc2 (will do rc3 over the wkd with some packaging fixes) as well.

I will certainly do, I'm migrating from 0.6.4 so I'm going step by step :)

README.md
adumesny: well I don't want to expose jquery dependencies as many might not use this lib otherwise, which is why we don't have an external dependency and all.js include everything you need to run. JQ was removed from 1.x API signature, and from the main code in 2.x, and eventually out and replaced with html5 drag&drop.

adumesny: Also is it not possible to have
```js
import 'gridstack/dist/gridstack.all.js'
```
anymore ? I thought that worked... and hides the internal usage of jq

adumesny: Also what about the jquery.js change I also had made
`
// Expose jQuery and $ identifiers, even in AMD
// (#7102#comment:10, https://github.com/jquery/jquery/pull/557)
// and CommonJS for browser emulators (#13566)
//if ( !noGlobal ) { [alain] noGlobal=true when using webpack/TS, yet we compile this in, not AMD load, so always expose
	window.jQuery = window.$ = jQuery;
//}
`

PR #1368, opened 2020-09-21
Add Vue.js example

tobias-grasse:
### Description
Added a basic example of how to integrate GridStack with a Vue.js app. This demo requires a modestly modern browser, but Vue has a legacy build mode when using single file components with Vue CLI. I added a few lines about common pitfalls when integrating a DOM-dependent library like gridstack.js with virtualDOM frameworks.

### Checklist

### DIFF
--- a/demo/vuejs.html
+++ b/demo/vuejs.html
@@ -1,0 +1,107 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Gridstack.js Vue integration example</title>
+    <link
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/npm/gridstack@2.0.0/dist/gridstack.min.css"
+    />
+
+    <script src="https://cdn.jsdelivr.net/npm/gridstack@2.0.0/dist/gridstack.all.js"></script>
+    <style>
+      .grid-stack {
+        background: #fafad2;
+      }
+      .grid-stack-item-content {
+        color: #2c3e50;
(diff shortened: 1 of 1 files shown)


demo/vuejs.html
```diff
@@ -0,0 +1,107 @@
...
+            }, 2000);
+          },
+        },
+        mounted: function () {
+          // Provides access to the GridStack instance across the Vue component.
+          this.grid = GridStack.init({ float: true });
+
+          // Use an arrow function so that `this` is bound to the Vue instance. Alternatively, use a custom Vue directive on the `.grid-stack` container element: https://vuejs.org/v2/guide/custom-directive.html
+          this.grid.on("dragstop", (event, element) => {
+            const node = event.target.gridstackNode;
```
adumesny, line 84: this might be just element.gridstackNode but haven't double checked.

demo/vuejs.html
```diff
@@ -0,0 +1,107 @@
...
+            // `this` will only access your Vue instance if you used an arrow function, otherwise `this` binds to window scope. see https://hacks.mozilla.org/2015/06/es6-in-depth-arrow-functions/
```
adumesny, line 85: thanks for all the comments. I haven't looked at what ES6 does these days, having been on Typescript for so long now...

adumesny (approved):
thanks for submitting this ! the only thing left to do, which I can take care of otherwise, is update index.html and the #gridstackjs-for-specific-frameworks readme section to point to your example.

demo/vuejs.html
tobias-grasse: You're right!

```suggestion
            const node = element.gridstackNode;
```

PR #1376, opened 2020-09-25
HTML5 DragDrop plugin implemented

rhlin:
### Description
1. I wrote a simple version of DDPlugin using HTML 5 dragdrop api without jquery.
2. add a demo (modified form Advanced) to test the plugin
3. add a static method to GridStack to obtains h5dd plugin for demo

### Checklist
- [ ] All tests passing (`yarn test`)   ---- 3  fail  because "resizable-disabed test by jquery( test case depends on jquery for resizable, i didn't use that)


src/gridstack.ts
```diff
@@ -21,6 +21,8 @@ export * from './gridstack-dd';
 // TEMPORARY import the jquery-ui drag&drop since we don't have alternative yet and don't expect users to create their own yet
```
adumesny, line 21: I guess I was wrong there! whoa...

src/gridstack.ts
```diff
@@ -87,6 +89,16 @@ export class GridStack {
...
     return el.gridstack
   }
+  /**
+   * Will return i-th DDPlusgin registerd in GridStackDD
+   * @param i i-th plugin (default to 0)
+   *
+   * @example
+   * let GridStackDDJQueryUI = GridStack.getDDPlugin(0);
+   */
+  public static getDDPlugin(i = 0) {
```
adumesny, line 99: I think a better solution now is to have the user explicitly include the one they want instead of loading them both - which will auto-register.

adumesny:
This is GREAT! I guess we might need to release 3.0 sooner than I thought. thank you so much for starting this html5 d&d version. I will take some time this weekend to go over it and try it out.

I might need to make some 2.x fixes before taking this in as it should be 3.x with the user choosing which D&D they want.

pwnyprod:
thats exactly what is needed for me <3 thank you so much

adumesny:
@pwnyprod | when will 3.0 released?

faster if I get help...

adumesny:
@rhlin I checked it in and now part of the build, but it's **very very sluggish** - wonder if you need to throttle some events (and use latest position, purge the rest) or what... I didn't have much time to play with it or check your code yet, but it's part of the build now (see #1388)

If you want to get on slack so we can communicate that might be best. I didn't know you were working on HTML5 D&D, but glad you did (I wasn't gonna start for a while).

rhlin:
I already throttle dragover event 100ms. It seems throttle doesn't get much perfomance improvement. 
The performance seems to be related to gpu or other standard configuration of computer .
I will try add `requestAnimateFrame` and css property `will-change`  next few day.

I left message to you on slack.

PR #1391, opened 2020-09-30
fixed animation issue by not recreating the entire stylesheet

marttop:
### Description
fixed animation issue by not recreating the entire stylesheet on every event but only when it's needed.

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -1072,13 +1072,12 @@ export class GridStack {
 
   /** @internal */
   private _initStyles(): GridStack {
-    if (this._stylesId) {
-      Utils.removeStylesheet(this._stylesId);
+    if (!this._stylesId) {
+      this._stylesId = 'gridstack-style-' + (Math.random() * 100000).toFixed();
+      // insert style to parent (instead of 'head' by default) to support WebComponent
+      let styleLocation = this.opts.styleInHead ? undefined : this.el.parentNode as HTMLElement;
+      this._styles = Utils.createStylesheet(this._stylesId, styleLocation);
     }
-    this._stylesId = 'gridstack-style-' + (Math.random() * 100000).toFixed();
-    // insert style to parent (instead of 'head' by default) to support WebComponent
-    let styleLocation = this.opts.styleInHead ? undefined : this.el.parentNode as HTMLElement;
-    this._styles = Utils.createStylesheet(this._stylesId, styleLocation);
     if (this._styles !== null) {
       this._styles._max = 0;
     }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -109,10 +109,8 @@ export class Utils {
   static getAllSelectors(sheet: CSSStyleSheet) { 
     var ret = [];
     var rules = sheet.cssRules;
-    // console.log(JSON.stringify(rules));
     for(var i = 0; i < rules.length; i++) {
       var rule = rules[i];
-      // console.log(JSON.stringify(rules));
         var index = rule.cssText.indexOf(" {");
         if(index >0){
           var sel = rule.cssText.substr(0,index);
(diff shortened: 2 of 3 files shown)


src/gridstack.ts
```diff
@@ -1072,13 +1072,12 @@ export class GridStack {
 
   /** @internal */
   private _initStyles(): GridStack {
-    if (this._stylesId) {
-      Utils.removeStylesheet(this._stylesId);
+    if (!this._stylesId) {
```
adumesny, line 1075: I would move that create code to first line of _updateStyles() and change constructor to call it instead. Remove _initStyles()

src/utils.ts
```diff
@@ -106,13 +106,42 @@ export class Utils {
     if (!el || !el.parentNode) return;
     el.parentNode.removeChild(el);
   }
+  static getAllSelectors(sheet: CSSStyleSheet) {
```
adumesny, line 109: return type, method description

src/utils.ts
```diff
@@ -106,13 +106,42 @@ export class Utils {
     if (!el || !el.parentNode) return;
     el.parentNode.removeChild(el);
   }
+  static getAllSelectors(sheet: CSSStyleSheet) { 
+    var ret = [];
+    var rules = sheet.cssRules;
+    // console.log(JSON.stringify(rules));
+    for(var i = 0; i < rules.length; i++) {
```
adumesny, line 112: for () spacing please, same with if ()

src/utils.ts
```diff
@@ -106,13 +106,42 @@ export class Utils {
...
+        var index = rule.cssText.indexOf(" {");
+        if(index >0){
+          var sel = rule.cssText.substr(0,index);
+          ret.push(sel);
+        }
+    }
+    return ret;
+}
 
+static hasCSSRule(sheet: CSSStyleSheet, selector: string) {
```
adumesny, line 123: same

src/utils.ts
```diff
@@ -106,13 +106,42 @@ export class Utils {
...
+    }
+    return false;
+}
   /** inserts a CSS rule */
   static addCSSRule(sheet: CSSStyleSheet, selector: string, rules: string) {
-    if (typeof sheet.addRule === 'function') {
-      sheet.addRule(selector, rules);
-    } else if (typeof sheet.insertRule === 'function') {
-      sheet.insertRule(`${selector}{${rules}}`);
+    if(!Utils.hasCSSRule(sheet, selector)){
```
adumesny, line 134: I actually don't know what we need to check for existing rules... I mean we know what the previous max was so we can append new styles only. Also the original one are created once so they don't need this.

src/gridstack.ts
```diff
@@ -1072,13 +1072,12 @@ export class GridStack {
...
+      // insert style to parent (instead of 'head' by default) to support WebComponent
+      let styleLocation = this.opts.styleInHead ? undefined : this.el.parentNode as HTMLElement;
+      this._styles = Utils.createStylesheet(this._stylesId, styleLocation);
     }
-    this._stylesId = 'gridstack-style-' + (Math.random() * 100000).toFixed();
-    // insert style to parent (instead of 'head' by default) to support WebComponent
-    let styleLocation = this.opts.styleInHead ? undefined : this.el.parentNode as HTMLElement;
-    this._styles = Utils.createStylesheet(this._stylesId, styleLocation);
     if (this._styles !== null) {
       this._styles._max = 0;
```
adumesny, line 1082: this should NEVER happen but on creation, then we can use the max to figure new styles that need to be appended.
you don't need to check for existing styles.

adumesny (changes requested):
thanks for doing this, but this is wrong approach to check for existing styles. we know what the prev max was and can append only new styles instead.
And yes we should never delete the file UNLESS there are some styles which are column based or something that can change like margin. I haven't looked carefully on what is inserted in there...

adumesny:
Also please combine all those into 1 commit please - you can `git rebase -i HEAD~6` (or how many commits you've made) and squash them into one and git push -f your changes up.

adumesny:
and looks like changing the `cellHeight()` or `margin()` needs to delete and re-create, and that's the only time, otherwise we append new style to initial creation that are past max (no need to check for existing).

Issue #937, opened 2018-08-16
Animate doesn't work


ermcgrat:
Animate doesn't work. You can try any* of the demos and add animate: true to them, and animation will not work. It doesn't matter whether you use the configuration object, data-gs-animate, or setAnimation() api.

*Adding animate to the two-grid demo does work. Animate works as long as you have two or more grids on your page, see details below.

## Your environment
Gridstack v 1.0.0-dev
Firefox, Chrome, Edge, IE11

## Steps to reproduce
I have taken the float demo and simply added animate: true. See here:
https://gridstackdemo.azurewebsites.net/demo/float.html

Note that while rearranging items they are not animated. The grid itself (height, etc) is animated, the item being dragged is animated (when released). Other items pushed out of the way (by dragging a block around) will not be animated.

## Expected behaviour
Animation should happen

## Actual behaviour
Items snap to position instead of animating.

## Additional Information
The animation isn't happening because all of the style information containing "top" and "height" properties gets deleted and recreated constantly.  See here (this method is called constantly while dragging a block around):
https://github.com/gridstack/gridstack.js/blob/7d9ac26847b89c593ab5ee5d778185091267d4d4/dist/gridstack.js#L1003

Since the animation is based upon CSS transition, it fails to work since the code is simply removing "top", etc. For example, what's essentially happening is a block goes through this:
top: 0px
-top property removed by stylesheet being deleted in javascript
-new stylesheet is created
top: 80px

Of course CSS can't transition between these states. Animation works as long as there are two or more grids because the other grid's stylesheet will still exist, and thus blocks will always have a top property instead of the interim undefined state. I would submit a pull request if I knew how to address this issue, but I'm at a loss why it is done this way.

My hacky workaround for the time being is to create my own stylesheet file with greater specificity so my blocks will always have some sort of top property.

nws-jstorm:
I just removed the line 1003 you mentioned in your post. Now the grid is behaving like it should, but i dont know what else is affected by this line. Did somebody the same and has experience with this?

radiolips:
@ermcgrat @nws-jstorm This is now fixed in the develop branch and will be deployed with the next release. Thanks for letting us know about the bug.

adumesny:
had to revert change until correct fix...

marttop:
Is there any update with this issue? "animate: true" still not working for me.
[...]

PR #1392, opened 2020-10-02
Fixed 'animate' issue.

marttop:
### Description
#937 Fixing this issue by not recreating the stylesheet.

#1391 I created another pull request (It's for sure not the proper way to do but I still struggle a lot with git) and followed @adumesny's advices :

- I don't check anymore for existing styles because I could prevent _max from being reset every time
For cellHeigh() and margin(), on method _updateStyles I added a boolean parameter 'doForceUpdate' witch is set to true only when called from cellHeight() or margin(). In that case I reset the stylesheet.
- I tried my best to respect your coding style as you asked.
- I tried to remove the _initStyles method and add it to _updateStyles but kept on breaking the behaviour so I let you do it... It seems I don't know enough on the component to do so.

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -255,7 +255,7 @@ export class GridStack {
 
     this._setStaticClass();
 
-    this._initStyles();
+    this._initStyles(true);
 
     this.engine = new GridStackEngine(this.opts.column, (cbNodes, removeDOM = true) => {
       let maxHeight = 0;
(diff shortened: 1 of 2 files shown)


adumesny:
thanks for taking a second look. I think I want to re-write this all thing rather than try to patch what's there... but I need to understand all corner cases first. We also should delete the style when destroy() happens (which we don't) so we need a separate removeStyle() method that will also reset the max info structure...

Probably best to let me do it at this point. I'm tied this weekend but might get to it middle of next week.

marttop:
I understand that you want to rewrite it, but just in case, this at least fixes the bug temporarily.

adumesny:
thank you for fix suggestion! I ended up re-writing that code instead #1395 AND default to animate since SO much nicer (I even didn't know about it).

PR #1395, opened 2020-10-04
fix to animate

adumesny:
### Description
* fix for #937, thank you @marttop for proposing a fix
* ended up re-writting the CSS creation routine and make it easier to maintain going forward.
* we now only re-create stylesheet when margin or cellHeight (the only static items) change.
Else we happen row based rules as the grids needs them (and only what it needs, not always +10)

### Checklist


marttop:
Thanks for this final fix!

PR #1404, opened 2020-10-06
Vue3  example

manang:
### Description
I've created an example of gridstack used in a vue3 js.


.gitignore
```diff
@@ -4,4 +4,5 @@
 .npmrc
 coverage
 dist
-node_modules
\ No newline at end of file
+node_modules
+.vscode
```
adumesny, line 8: please revert that

adumesny (changes requested):
Angelo, thank you for doing this, but can you make it EXACTLY the same as the 2.0 version that is checked in (down to the style and includes of demo styles) other than `import { createApp }` and `.mount("#app")` part so it's easier to diff the two and see what changed for 3.0 ?

Also, please append your file to index.html and readme link, maybe rename the existing one vue2ls.html

and finally, merge all your git changes into just 1 if possible (git rebase -i HEAD~3 and squash, the git push -f to update your commits), otherwise I can do it when I accept I suppose.

.gitignore
manang: why you want .vscode inside the repo?

adumesny: yeah you're right. never mind

PR #1412, opened 2020-10-16
Update gridstack.jQueryUI.js

leontb99:
### Description
Change to fix error mentioned in this thread: https://github.com/gridstack/gridstack.js/issues/858

### DIFF
--- a/src/gridstack.jQueryUI.js
+++ b/src/gridstack.jQueryUI.js
@@ -31,7 +31,7 @@
   JQueryUIGridStackDragDropPlugin.prototype.resizable = function(el, opts) {
     el = $(el);
     if (opts === 'disable' || opts === 'enable' || opts === 'destroy') {
-      el.resizable(opts);
+      el.resizable().resizable(opts);
     } else if (opts === 'option') {
       var key = arguments[2];
       var value = arguments[3];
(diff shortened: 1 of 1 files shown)


src/gridstack.jQueryUI.js
```diff
@@ -31,7 +31,7 @@
   JQueryUIGridStackDragDropPlugin.prototype.resizable = function(el, opts) {
     el = $(el);
     if (opts === 'disable' || opts === 'enable' || opts === 'destroy') {
-      el.resizable(opts);
+      el.resizable().resizable(opts);
```
adumesny, line 34: thank you for the fix! but it has the side effect of enabling resizable before destroying it, which is wasteful if it wasn't created to start with. looking to see if we can avoid this...

adumesny:
actually you are looking at an old code base... in 2.0 (typescript) you can see in https://github.com/gridstack/gridstack.js/blob/develop/src/jq/gridstack-dd-jqueryui.ts#L32
that we now have
```js
    } else if (opts === 'destroy') {
      if ($el.data('ui-draggable')) { // error to call destroy if not there
        $el.draggable(opts);
      }
```
so no need for your fix.

PR #1441, opened 2020-11-05
Bug fix: Init with marginTop prop or other margin props will turn height to 1 when resize a grid

deadivan:
### Description
Bug fix: Init with marginTop prop or other margin props will turn height to 1 when resize a grid

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -1854,6 +1854,10 @@ export class GridStack {
       this.opts.marginLeft = data.height;
       delete this.opts.margin;
     }
+    
+    //without this.opts.margin, getMargin method will return undefined, and affect resize behavior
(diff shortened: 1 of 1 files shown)


src/gridstack.ts
```diff
@@ -1854,6 +1854,10 @@ export class GridStack {
       this.opts.marginLeft = data.height;
       delete this.opts.margin;
     }
+    
+    //without this.opts.margin, getMargin method will return undefined, and affect resize behavior
+    if(!this.opts.margin) this.opts.margin = Math.round((this.opts.marginTop + this.opts.marginBottom) / 2);
```
adumesny, line 1859: this isn't the right fix because margin=0 is a valid value. Thanks for ringing this issue, will fix it the proper way next...

adumesny:
thanks for the suggestion! but ended up checking a correct fix.

PR #1474, opened 2020-11-21
Fix placeholder not carrying base item class

pablosichert:
Hey there,

this PR fixes a bug when using the `itemClass` option and dragging an item. The placeholder also needs the default `itemClass` (to be consistent with the behavior of items placed on the grid).

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -298,7 +298,7 @@ export class GridStack {
     placeholderChild.className = 'placeholder-content';
     placeholderChild.innerHTML = this.opts.placeholderText;
     this.placeholder = document.createElement('div');
-    this.placeholder.classList.add(this.opts.placeholderClass, this.opts.itemClass);
+    this.placeholder.classList.add(this.opts.placeholderClass, defaults.itemClass, this.opts.itemClass);
     this.placeholder.appendChild(placeholderChild);
 
     this._updateContainerHeight();


adumesny:
thanks for the fix. I supposed if you set opts.itemClass to something other than `grid-stack-item` we still need grid-stack-item in the DOM as the CSS is still dependent on that. I think `addWidth()` will also need to add this.ops.itemClass  to this code

`doc.body.innerHTML = `<div class="grid-stack-item"><div class="grid-stack-item-content">${content}</div></div>`;`

if you don't mind doing that.

pablosichert:
[addWidget](https://github.com/dashql/gridstack.js/blob/21474391ba6669760a9577ef8d757a3d33eccf29/src/gridstack.ts#L329)

[calls](https://github.com/dashql/gridstack.js/blob/21474391ba6669760a9577ef8d757a3d33eccf29/src/gridstack.ts#L372)

[makeWidget](https://github.com/dashql/gridstack.js/blob/21474391ba6669760a9577ef8d757a3d33eccf29/src/gridstack.ts#L741)

[calls](https://github.com/dashql/gridstack.js/blob/21474391ba6669760a9577ef8d757a3d33eccf29/src/gridstack.ts#L743)

[_prepareElement](https://github.com/dashql/gridstack.js/blob/21474391ba6669760a9577ef8d757a3d33eccf29/src/gridstack.ts#L1432)

which [sets this.opts.itemClass](https://github.com/dashql/gridstack.js/blob/21474391ba6669760a9577ef8d757a3d33eccf29/src/gridstack.ts#L1433).

So I think it should be fine now without further modifications.

PR #1476, opened 2020-11-22
Add react example

eloparco:
### Description
Add React usage example (replicating the behavior/application in Vue-gridstack.js examples).


adumesny:
this is great! thank you so much... might need to do an angular now (not sure we can html inline like you did as usually have entire typescript projects.
I'll update the doc to point to this!

PR #1492, opened 2020-11-29
'data-gs-min-width' to 'gs-min-w' attr rename

adumesny:
### Description
* removed 'data-' from all attributes (legacy JS style) and shorten width/height to w/h
* make for smaller and more efficient code and CSS.
Easy enough to global replace in your code.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Join us on Slack: https://gridstackjs.troolee.com
   - [Migrating to v0.6](#migrating-to-v06)
   - [Migrating to v1](#migrating-to-v1)
   - [Migrating to v2](#migrating-to-v2)
+  - [Migrating to v3](#migrating-to-v3)
 - [jQuery Application](#jquery-application)
 - [Changes](#changes)
 - [The Team](#the-team)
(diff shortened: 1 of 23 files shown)


flawedvice:
I'm using gridstack.js along with TypeScript and SvelteKit, but if I try to add custom attributes like `gs-w`, TypeScript complains:

Screenshot from 2023-02-13 12-24-33

It works fine though, but the error is annoying.

I also noticed that the error disappears when prefixing the attribute with `data-`, as it's a "valid HTML attribute" for TS.

Is there anything one can do to bypass this?

adumesny:
@flawedvice I don't know sveltekit but this works in regular html as documented `<div class="grid-stack-item" gs-w="2">` and for the Angular framework this is what I use `[attr.gs-w]="2"`

flawedvice:
Surely it has to do with Svelte's types for HTML elements. What actually worked for me, and hopefully for others who might have the same problem, is to use the spread operator over the attributes:

image

JarolinVargas:
Best practice is to include data- prefix, as that's what's required when adding custom attributes to HTML elements. https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#custom-data-attribute

adumesny:
yes I know that now (see https://github.com/gridstack/gridstack.js/pull/2904) but wondering if we can't just use `[attr.gs-w]="2"` like I'm using in Angular to let it know it's an attribute ?

Issue #1001, opened 2019-07-18
Can't resize from bottom handler when the grid item is at the bottom of page/container


jonathangriguol:
##Can't resize from bottom handler when the grid item is at the bottom of page/container

I want to resize from the bottom border a Grid Item which is placed at the bottom of the Grid container (and screen o parent of this grid container)
Even the grid is initialized with height = 0 and draggable.scroll = true the item collides with the lower edge and the Grid doesn't increase it's heigth size
I try to override resizable.resize method, but I can't, also can't listen on resize event for each item grid in order to increase manually the grid heigth.

Example from gridstack site (http://gridstackjs.com/demo/float.html)
red arrows indicates that I'm trying to resize on this direction frmo the lower edge.

![image](https://user-images.githubusercontent.com/13875561/61487136-7e818600-a97b-11e9-92c4-9badb1d26c50.png)

Can help me? thanks!

## Your environment
* gridstack.js 1.0.0-dev (any version)
* Version 75.0.3770.80 (Official Build) (64-bit)

## Steps to reproduce
Test here:
https://jsfiddle.net/td6zyL8v/1/
Try to resize the 3rd item more than the visible height

## Expected behaviour
I think the grid container should auto scroll and increase its heigth, like if you are dragging a item grid

## Actual behaviour
You can't resize more than the bottom limit

adumesny:
I'm not seeing this at all (Chrome 79, and https://gridstackjs.com/demo/float.html
I can drag below OR resize below and it will auto-scroll while doing it just fine.
closing for now unless you can show video issue.

hbcarlos:
Hi, I've been trying v3.1 and is working smoothly. Thanks!

The drag scrolling issue is solved but unfortunately, the resize scrolling issue still there. You can see the issue in the example https://gridstackjs.com/demo/float.html. If you try to resize at the bottom of the screen it doesn't work but does the drag and drop.

![Screen Recording 2020-12-05 at 22 54 06](https://user-images.githubusercontent.com/26092748/101266301-d9bd3f80-374d-11eb-8ac8-376ba6b10062.gif)

adumesny:
because items will not grow unless half way past mid point of a grid height, the only way to fix that is it add an additional extra bottom row when item you are currently dragging/sizing is near the bottom, then make it normal height when done...

hbcarlos:
Thanks for the advice. I've been trying what you said but I'm having some troubles.

[...]

PR #1530, opened 2020-12-09
Resizing issue

hbcarlos:
### Description

As discussed in #1001 the resizing with scrolling is not working.
I'm trying to solve it here but I'm having some troubles. Now is working fine but I'm resizing the layout adding some rows on `resizestart` because when while resizing an item there is a back and force with the number of cells creating a strange behaviour.

### Checklist

### DIFF
--- a/src/gridstack-dd.ts
+++ b/src/gridstack-dd.ts
@@ -342,7 +342,6 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
   // variables used/cashed between the 3 start/move/end methods, in addition to node passed above
   let cellWidth: number;
   let cellHeight: number;
-  let minRow: number;
 
   /** called when item starts moving/resizing */
   let onStartMoving = (event: Event, ui: DDUIData): void => {
@@ -352,9 +351,6 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
     if (this._gsEventHandler[event.type]) {
       this._gsEventHandler[event.type](event, target);
     }
-
-    // Saving `minRow`
-    minRow = this.opts.minRow;
 
     this.engine.cleanNodes();
     this.engine.beginUpdate(node);
@@ -426,17 +422,8 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
       if (node._lastTriedX === x && node._lastTriedY === y) return;
     } else if (event.type === 'resize')  {
       if (x < 0) return;
-      // Addin an extra row if the item it's at the bottom of the layout
-      if (node.y+node.h >= this.getRow()-1) {
-        this.opts.minRow = this.getRow() + 1;
-        this._updateContainerHeight();
-      }
-
       // Scrolling page if needed
       Utils.updateScrollResize(event as MouseEvent, el, cellHeight);
-      // Restore minRow
-      this.opts.minRow = minRow;
-
       w = Math.round(ui.size.width / cellWidth);
       h = Math.round(ui.size.height / cellHeight);
       if (w === node.w && h === node.h) return;
--- a/src/gridstack-dd.ts
+++ b/src/gridstack-dd.ts
@@ -353,10 +353,8 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
       this._gsEventHandler[event.type](event, target);
     }
 
-    // Saving `minRow` and adding new ones for resizing
+    // Saving `minRow`
     minRow = this.opts.minRow;
-    this.opts.minRow = this.getRow() + 100;
-    this._updateContainerHeight();
 
     this.engine.cleanNodes();
     this.engine.beginUpdate(node);
@@ -428,8 +426,17 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
       if (node._lastTriedX === x && node._lastTriedY === y) return;
     } else if (event.type === 'resize')  {
       if (x < 0) return;
+      // Addin an extra row if the item it's at the bottom of the layout
+      if (node.y+node.h >= this.getRow()-1) {
+        this.opts.minRow = this.getRow() + 1;
+        this._updateContainerHeight();
+      }
+
       // Scrolling page if needed
       Utils.updateScrollResize(event as MouseEvent, el, cellHeight);
+      // Restore minRow
+      this.opts.minRow = minRow;
+
       w = Math.round(ui.size.width / cellWidth);
       h = Math.round(ui.size.height / cellHeight);
       if (w === node.w && h === node.h) return;
@@ -490,10 +497,6 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
     this._triggerChangeEvent();
 
     this.engine.endUpdate();
-
-    // Removing lines rows added on `resizestart`
-    this.opts.minRow = minRow;
-    this._updateContainerHeight();
 
     // if we re-sized a nested grid item, let the children resize as well
     if (event.type === 'resizestop') {
--- a/src/gridstack-dd.ts
+++ b/src/gridstack-dd.ts
@@ -342,6 +342,7 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
   // variables used/cashed between the 3 start/move/end methods, in addition to node passed above
   let cellWidth: number;
   let cellHeight: number;
+  let minRow: number;
 
   /** called when item starts moving/resizing */
   let onStartMoving = (event: Event, ui: DDUIData): void => {
@@ -351,6 +352,11 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
     if (this._gsEventHandler[event.type]) {
       this._gsEventHandler[event.type](event, target);
     }
+
+    // Saving `minRow` and adding new ones for resizing
+    minRow = this.opts.minRow;
+    this.opts.minRow = this.getRow() + 100;
+    this._updateContainerHeight();
 
     this.engine.cleanNodes();
     this.engine.beginUpdate(node);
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -327,5 +327,27 @@ export class Utils {
       }
     }
   }
+
+  /** @internal */
+  static updateScrollResize(event: MouseEvent, el: HTMLElement, distance: number): void {
+    let scrollEl = this.getScrollParent(el);
+    if (scrollEl === null) return;
```
adumesny, line 334: if (!scrollEl ) return;

src/utils.ts
```diff
@@ -327,5 +327,27 @@ export class Utils {
...
+    const top = event.clientY < distance;
+    const bottom = event.clientY > height - distance;
+
+    if (top) {
+      // This also can be done with a timeout to keep scrolling while the mouse is
+      // in the scrolling zone. (will have smoother behavior)
+      scrollEl.scrollBy({ behavior: 'smooth', top: event.clientY - distance});
+    }
+
+    if (bottom) {
```
adumesny, line 348: else if (bottom)

src/gridstack-dd.ts
```diff
@@ -352,6 +353,11 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
       this._gsEventHandler[event.type](event, target);
     }
 
+    // Saving `minRow` and adding new ones for resizing
+    minRow = this.opts.minRow;
+    this.opts.minRow = this.getRow() + 100;
```
adumesny, line 358: big no no - don't arbitrary add 100 rows!
we should only add 1 extra row (when computing the needed rows) IFF the item you are resizing/moving is at the bottom (node.y+node.h). don't need to mod minRow for this...

src/utils.ts
```diff
@@ -327,5 +327,27 @@ export class Utils {
       }
     }
   }
+
+  /** @internal */
```
adumesny, line 331: put a real comment on your code. seems reasonable given we do something similar (though quite different code. I will have to diff into the scroll as I didn't write that)

src/gridstack-dd.ts
```diff
@@ -422,6 +428,8 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
       if (node._lastTriedX === x && node._lastTriedY === y) return;
     } else if (event.type === 'resize')  {
       if (x < 0) return;
+      // Scrolling page if needed
+      Utils.updateScrollResize(event as MouseEvent, el, cellHeight);
```
adumesny, line 432: ok

src/h5/dd-resizable.ts
```diff
@@ -37,6 +38,8 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal */
   private temporalRect: Rect;
   /** @internal */
+  private scrollY: number;
```
adumesny, line 41: do we need that in JQ version as well ?

src/h5/dd-resizable.ts
```diff
@@ -152,6 +155,8 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal */
   private _resizeStart(event: MouseEvent): DDResizable {
     this.originalRect = this.el.getBoundingClientRect();
+    const scrollEl = Utils.getScrollParent(this.el);
+    this.scrollY = scrollEl === null ? 0 : scrollEl.scrollTop;
```
adumesny, line 159: this.scrollY = scrollEl ? scrollEl.scrollTop : 0;

src/h5/dd-resizable.ts
```diff
@@ -216,12 +222,15 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal */
   private _getChange(event: MouseEvent, dir: string): Rect {
     const oEvent = this.startEvent;
+    const scrollEl = Utils.getScrollParent(this.el);
+    const scrolled = scrollEl === null ? 0 : (scrollEl.scrollTop - this.scrollY);
```
adumesny, line 226: similar change

src/h5/dd-resizable.ts
```diff
@@ -152,6 +155,8 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal */
   private _resizeStart(event: MouseEvent): DDResizable {
     this.originalRect = this.el.getBoundingClientRect();
+    const scrollEl = Utils.getScrollParent(this.el);
```
adumesny, line 158: cache this since you need it in all 3 methods and continuously, or does it change over time ?

adumesny (commented):
thank you for doing this! but minRow + 100 is a big no. Ping me if you need more help on fixing right way.

src/h5/dd-resizable.ts
hbcarlos: This changes over time. When there is no scroll because the layout is small `scrollEl` is null and eventually when the layout grows returns the scrollable element.

hbcarlos:
Thanks for the tips @adumesny. I need some help with the placeholder and the view of the item while resizing.

There is two case that I'm not sure how to solve. 
As you can see in the next gif, when the scrolling starts the item position changes.
item-move

When the item is at the bottom and the mouse pointer goes out of the `grid-stack` element while resizing, the `Utils.getScrollParent(this.el);` returns the `grid-stack` element and not his parent.

item-disapears

src/h5/dd-resizable.ts
adumesny: right, so you could `scrollEl = scrollEl || Utils.getScrollParent(this.el)` and cache it once it appears, unless it changes values ?
quick looking it appears to go up recursively so maybe it could be different elements over time...

adumesny:
I'll take a look in the next couple days... I just spend entire weekend re-writing collision so need to take a break from gridstack

hbcarlos:
The resize is working. The problem was the scrollable parent, sometimes got lost.

resize

adumesny:
sorry, taking a look now. I had to fix some Safari mac issues and released I also tweaked on we do resizing (event handling really).

src/utils.ts
```diff
@@ -327,5 +327,31 @@ export class Utils {
...
+   * 
+   * @param event `MouseEvent` that triggers the resize
+   * @param el `HTMLElement` that's being resized
+   * @param distance Distance to scroll
+   */
+  static updateScrollResize(event: MouseEvent, el: HTMLElement, distance: number): void {
+    const scrollEl = this.getScrollParent(el);
+    const height = scrollEl.clientHeight;
+
+    const top = event.clientY < distance;
```
adumesny, line 344: so you're gonna scroll if if mouse is within 'cellHeight' of the edges... might want to document that better, including the param info.

src/h5/dd-resizable.ts
```diff
@@ -152,6 +159,8 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal */
   private _resizeStart(event: MouseEvent): DDResizable {
     this.originalRect = this.el.getBoundingClientRect();
+    this.scrollEl = Utils.getScrollParent(this.el);
```
adumesny, line 162: reset this in destroy()

adumesny (commented):
that looks reasonable (minor doc comments). Can you update to the latest dev code so I can push it in ?
I need to cherry pick as well to verify it works. Have you tried how tat affects the JQ version as well ?

hbcarlos:
> sorry, taking a look now. I had to fix some Safari mac issues and released I also tweaked on we do resizing (event handling really).

No problem at all.

Thanks for merging and sorry for the comments, I can open a PR to add some comments if you want.

src/utils.ts
```diff
@@ -280,15 +280,15 @@ export class Utils {
 
   /** @internal */
   static getScrollParent(el: HTMLElement): HTMLElement {
-    let returnEl;
-    if (el === null) {
-      returnEl = null;
-    } else if (el.scrollHeight > el.clientHeight) {
-      returnEl = el;
+    if (el === null) return document.documentElement;
+    const style = getComputedStyle(el);
```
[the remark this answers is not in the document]
hbcarlos: Yes, also with `el.scrollHeight > el.clientHeight` sometimes is selecting a non scrollable parent.

src/utils.ts
hbcarlos: Yes, I check if the mouse is a certain distance to the edges (cellHeight) to start scrolling. Also, the closer to the edges the larger the scroll.

adumesny:
no need, I took care of it I believe. thanks for fixing this issue!

src/utils.ts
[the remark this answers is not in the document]
adumesny: interesting. didn't think that would be possible... but I realize the call you have return the first that could scroll which is safer anyway... but always return the doc as least, which might not scroll so it should return NULL and have that be handled ideally...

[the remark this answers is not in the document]
hbcarlos: Exactly, the best option would be returning `NULL` when there isn't a scrollable parent but in this case, we need to prevent `gridstack` to continue growing.

I was using the example `float.html` to check my changes. I think in this example the ideal would be that `gridstack` stops growing when there is no more page because the element that is scrolling is the document not the container that wraps `gridstack`.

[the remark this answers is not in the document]
hbcarlos: And we need to manually prevent `gridstack` to grow because:
In my case, I have two screens one on top of another and I have the browser with the webpage with gridstack on the screen above.
Captura de pantalla 2021-01-03 a las 8 24 12

When I try to resize or drag and drop an item to the bottom my mouse continue on the screen bellow so `gridstack´ still growing even when the page can't scroll. Once I stop there is no scrollable parent so the `documentElement` is resized to wrap everything.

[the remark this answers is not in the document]
adumesny: well we have a `maxRow` to prevent grid from growing past that size, but you are saying that if none of the parents can scroll (by default divs will auto scroll unless you have `overflow: hidden` or something) then we should stop growing so we don't clip, which makes sense... Looks like you are checking for overflow | overflow-y for and auto or scroll
you don't need 2 screens to test this. browser smaller than full screen will do that too...

if you want to take a stab at it that would be great. `_prepareDragDropByNode()` should cache if we can not v-scroll and prevent move/size past the max somehow.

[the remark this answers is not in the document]
hbcarlos: I'll take a look at this over the week (probably weekend, I'm quite busy right now) :)

Issue #735, opened 2017-08-08
anyone has reactjs using gristack example


eagoo:
while https://github.com/pitrho/react-gridstack is not working

DoctaWorm:
As much as I like gridstack and have been using it for about two years, if you're in a pure react environment I'd recommend looking into this one: https://github.com/STRML/react-grid-layout

derwaldgeist:
Unfortunately, react-grid-layout doesn't work well on mobile (touch) devices. Is there another alternative? I also like gridstack, but the (fixed) dependencies on jQuery and jQueryUI are a problem for me.

radiolips:
@derwaldgeist The dependencies will be removed in version 1. This has been delayed for no particularly good reason, but the dependencies _will_ be removed. The resize/dragging functionality will need to be supplemented (I've only been working with one other plugin but any plugin should work - you'll also be able to roll your own if you'd like).

@eagoo shoot. I'm sorry. I think there are a few other react tickets open, but I'll leave this open for now. If anyone (anyone!) wants to participate in getting gridstack and react working in harmony, please do so. I'd be happy to link a working example it in the readme.

jtheberge:
@eagoo I'm currently using React with this right now, and I prefer it over react-grid-layout because it uses absolute positioning rather than transformations to move things around, meaning any children with fixed positions aren't thrown off position.

The way you want to set it up is by letting React mount your initial widget(s) to the DOM first through its render, then with the lifecycle method componentDidMount(), initialize your grid with gridstack() and targetting where you want it initialized. 
Example: $(<grid element>).gridstack().

After each time you are adding a widget with React, make sure to then use the API method makeWidget  inside React's lifeycle method componentDidUpdate(), which makes the widget part of the grid directly after React mounts the new widget to the DOM.
Example:  $(<grid element>).data('gridstack').makeWidget($(<element to make>);

When you remove a widget with React, you must first remove it from the grid system with the API method removeWidget. Use the React lifecycle method componentWillUpdate() for this to ensure that you properly remove the item from the grid system before React deletes it from the DOM.
Example:  $(<grid element>).data('gridstack').removeWidget($(<element to remove>));

I hope this was helpful for anyone wishing to incorporate this great plugin with React.

@radiolips I don't have an easy to read example currently, but I might be able to come up with something in the future if you need it. The setup shouldn't be too complex.

ramigg:
[...]

PR #1531, opened 2020-12-10
Added example using a controlled React component and hooks. #735

zachsa:
### Description
Added an example do the demos folder using React hooks and a controlled, functional component.

### DIFF
--- a/demo/react-hooks.html
+++ b/demo/react-hooks.html
@@ -1,0 +1,94 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Gridstack.js React integration example</title>
+  <link rel="stylesheet" href="demo.css" />
+  <script src="../dist/gridstack-h5.js"></script>
+
+  <!-- Scripts to use react inside html -->
+  <script src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
+  <script src="https://unpkg.com/react-dom@16/umd/react-dom.production.min.js"></script>
+  <script src="https://unpkg.com/babel-standalone@6.15.0/babel.min.js"></script>
+</head>
+
+<body>
+  <div id="root"></div>
+</body>
+
+<script type="text/babel">
+  const { useState, useEffect, createRef, useRef } = React
+
+  const Item = ({ id }) => <div>I am item: {id}</div>
+
+  const ControlledStack = ({ items, addItem }) => {
+    const refs = useRef({})
+
+    if (Object.keys(refs.current).length !== items.length) {
+      items.forEach(({ id }) => {
+        refs.current[id] = refs.current[id] || createRef()
+      })
+    }
+
+    useEffect(() => {
+      const grid = GridStack.init()
+      grid.removeAll(false)
+      items.forEach(({ id }) => {
+        grid.makeWidget(refs.current[id].current)
+      })
+    }, [items])
+
+    return (
+      <div>
+        <button onClick={addItem}>Add new widget</button>
+        <div className="grid-stack">
+          {items.map((item, i) => {
+            return (
+              <div
+                ref={refs.current[item.id]}
+                key={item.id}
+                className={'grid-stack-item'}
+                gs-w="12"
+              >
+                <div className="grid-stack-item-content">
+                  <Item {...item} />
+                </div>
+              </div>
+            )
+          })}
+        </div>
+      </div>
+    )
+  }
+
+  const App = props => {
+    const [items, setItems] = useState([
+      { id: 'item-1' },
+      { id: 'item-2' }
+    ])
+
+    return (
+      <div>
+        <h1>Using GridStack.js with controlled React component and hooks</h1>
+        <p>
+          As with any virtualDOM-based framework, you need to check if React
+              has rendered the DOM (or any updates to it){" "}
+          <strong>before</strong> you initialize GridStack or call its
+              methods. This example shows how to make rendered components widgets:
+          </p>
+        <ol>
+          <li>Render items, each with a reference</li>
+          <li>Convert each rendered item to a widget using the reference and the  <a href="https://github.com/gridstack/gridstack.js/tree/develop/doc#makewidgetel">
+            makeWidget</a> function</li>
+        </ol>
+        <ControlledStack items={items} addItem={() => setItems([...items, { id: `item-${items.length + 1}` }])} />
+      </div>
+    )
+  }
+
+  ReactDOM.render(<App />, document.getElementById("root"))
+</script>
+
+</html>
--- a/demo/react-hooks.html
+++ b/demo/react-hooks.html
@@ -15,7 +15,28 @@
 </head>
 
 <body>
-  <div id="root"></div>
+  <div>
+    <h1>Using GridStack.js with React hooks</h1>
+    <p>
+      As with any virtualDOM-based framework, you need to check if Reacthas rendered the DOM (or any updates to it)
+      <strong>before</strong> you initialize GridStack or call its methods. This example shows how to make rendered
+      components widgets:
+    </p>
+    <ol>
+      <li>Render items, each with a reference</li>
+      <li>Convert each rendered item to a widget using the reference and the <a
+          href="https://github.com/gridstack/gridstack.js/tree/develop/doc#makewidgetel">
+          makeWidget</a> function</li>
+    </ol>
+  </div>
+  <div>
+    <h2>Controlled stack</h2>
+    <div id="controlled-stack"></div>
+  </div>
+  <div>
+    <h2>Uncontrolled stack</h2>
+    <div id="uncontrolled-stack"></div>
+  </div>
 </body>
 
 <script type="text/babel">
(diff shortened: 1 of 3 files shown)


demo/react-hooks.html
```diff
@@ -0,0 +1,94 @@
...
+    const refs = useRef({})
+
+    if (Object.keys(refs.current).length !== items.length) {
+      items.forEach(({ id }) => {
+        refs.current[id] = refs.current[id] || createRef()
+      })
+    }
+
+    useEffect(() => {
+      const grid = GridStack.init()
```
adumesny, line 36: are you calling init() multiple times ? not the right thing. call once and save the var instead. Also removing all then making widgets doesn't seem right. the Vue example does only on what got added...

demo/react-hooks.html
```diff
@@ -0,0 +1,94 @@
...
+      <div>
+        <button onClick={addItem}>Add new widget</button>
+        <div className="grid-stack">
+          {items.map((item, i) => {
+            return (
+              <div
+                ref={refs.current[item.id]}
+                key={item.id}
+                className={'grid-stack-item'}
+                gs-w="12"
```
adumesny, line 53: don't think you want entire row widget

demo/react-hooks.html
zachsa: Okay. Easy to call init() just once. 

Looking at the Vue example I see that there is a button that adds an additional widget. However in my case that wouldn't work, as the component receives a list of items from a server that I would like to render as widgets. When I update the server, the component is re-rendered with an updated list of items that I would like to make into widgets. That is why I am using `grid.makeWidget(someRef.current)`

demo/react-hooks.html
zachsa: My intention here is that as a user adds items to the array (i.e. updates the database), that these new items come through to the client and get added to the bottom of the grid stack as rows. Users can then re-adjust the layout to incorporate those rows.

Is there a better way of achieving this?

zachsa: Ah, I see that I would need to recreate the grid from some kind of state, otherwise grid items will always be full rows on first component render

demo/react-hooks.html
adumesny: if you're reading a list from a server, I would call `load()` instead of makeWidget(). ha but wait the DOM elements are already all there ? then I would wrap the loop with a batchUpdate()/commit() so layout is only done once...

demo/react-hooks.html
adumesny: you can serialize the x,y,w,h of each grid item. adding them to the bottom and making when a full row is just weird. Look at view adding a single widget but you say yours come from backend as complete list. Which may or may not be what you want in general. I would suggest showing both cases.

zachsa: I've updated the example to include the old react example, re-written using functional components and also the controlled example

demo/react-hooks.html
zachsa: Thanks - works. Also, I see that NOT specifying the row in the JSX that there is no bug regarding errant order changes (the bug that I mentioned earlier)

adumesny:
thanks, that looks reasonable. would it make sense to replace the other one instead with this (or rename it as it maybe primary use case ?

zachsa:
I think both are useful as references. I'm not sure which would be considered the main example, but I would easily be able to find react code that I find useful from these two files as they are named. 

Also happy to make any further adjustments if you have any preference or ideas. 

Thank you for merging!

adumesny:
I didn't run it/try it, but adding a single widget should NOT call removeAll() + makeWidget() - only the newly added widget should have a makeWidget() call like the Vue example. That's really inefficient and not a good example.

zachsa:
I'm using that in my work - it seems to work okay for the use case. The reason is that I also need to remove widgets, but the removed widgets are not present in the array when the component renders.

It's definitely possible to keep track of widgets that have been added/removed, but calling removeAll() is vastly simpler and unless there is an obvious performance problem the tradeoff is definitely worth it.

A much better option would be to call `grid.isWidget(el)`. Could gridstack keep a dictionary of added elements somewhere? the key could be the hash of the the html string, and then every element turned into a widget just needs a unique id.

I agree that this is not necessary for the example though, since no elements are ever removed. (but the tricky part is all about removing elements, not adding them)

adumesny:
`grid.isWidget(el)` is same as checking `el.gridstackNode != undefined`, and we keep track of nodes (grid.engine.nodes) which have node.el HTML element back so if the item has been removed from DOM grid.engine.nodes[i].el.parentNode == null and you could call removeWidget() on it. You can also store an id with each widget, to look things up (we also internally store _id that is guarantee to be unique for lookup and layouts). But you still need to track newly added (what the Vue.js examples do).

yeah if you have no easy way to track newly added (removed can be done as shown) and want to call on a list of items, then yes remove all and re-create will work abeit not very efficiently and hopefully not leaking memory along the way :)

jedwards1211:
I just got to where I can drag an external widget onto my grid with React, but it's kind of awkward.

GridStack has already added the new DOM element, but it's out of React's control (it's best if you can avoid having any DOM state outside of React's control).  And when I add the new item to local state, React is going to render an element for it and call `makeWidget` on it -- so I have to `removeWidget` on the element GridStack added for the drop, lest two copies of the new element appear.

I also would like to make the cloned external widget being dragged a live React component (e.g. when passing a `helper` function to `initDragIn`).  Maybe it could be done with [`ReactDOM.createPortal`](https://reactjs.org/docs/portals.html)?  But of course it wouldn't ever get my grid component's enclosing React context; only the new widget created by my grid component would get its enclosing React context.

In general the two-way binding nature of GridStack and the one-way dataflow philosophy of React are in conflict, and the compatibility is more tenuous than the docs let on.  I wish GridStack had a mode where it gives me change events without doing any DOM manipulation at all, and leaves it up to me to accomplish the DOM manipulations.

demo/react-hooks.html
jedwards1211: @adumesny if you need to have live React components, using `load()` isn't a very good option.  I guess you could render React components into the widgets using [ReactDOM.createPortal](https://reactjs.org/docs/portals.html) after `load()` creates the elements, but then you would still need other code to hook up widgets added later.

adumesny:
| I wish GridStack had a mode where it gives me change events without doing any DOM manipulation at all, and leaves it up to me to accomplish the DOM manipulations.

@jedwards1211 please file this as a feature request with a good reason/example showing why it would be better for certain frameworks.

Also if you can improve the existing React example for others, that would be great.

jedwards1211:
@adumesny

> adding a single widget should NOT call removeAll() + makeWidget()

The only options for handling changes to the `items` array in a controlled React component are
- do removeAll() + makeWidget()
- compare the new item array with the DOM elements that are currently in the grid and makeWidget() / removeWidget() as necessary to bring the grid up to date

removeAll() + makeWidget() is less performant but has the virtue of simplicity.  It also takes care of the issue I was talking about in my previous comment where I have to `removeWidget` on a newly dropped item.

It's fairly idiomatic anyway since in React, for better or worse, re-rendering a list after adding or removing an item involves creating O(n) new virtual DOM elements, which React then compares to the current virtual DOM, another O(n) operation.  If only one element changed it will only make one update to the actual DOM.  The main exception to this is virtualization libraries like `react-virtualized` or `react-window` that only render elements within the current scroll viewport of a list into the DOM.

I had made my own code call `removeWidget` when a single item I rendered unmounts, but this approach ended up a failure -- by this time React has already removed the element from the DOM so `removeWidget` ignores it.

v1talii-dev:
> I just got to where I can drag an external widget onto my grid with React, but it's kind of awkward.
> 
> GridStack has already added the new DOM element to the grid, but it's out of React's control (it's best if you can avoid having any DOM state outside of React's control). And when I add the new item to local state, my React component is going to render an element for it and call `makeWidget` on it -- so I have to `removeWidget` on the element GridStack added for the drop, lest two copies of the new element appear.
> 
> I also would like to make the cloned external widget being dragged a live React component (e.g. when passing a `helper` function to `initDragIn`). Maybe it could be done with [`ReactDOM.createPortal`](https://reactjs.org/docs/portals.html)? But of course it wouldn't ever get my grid component's enclosing React context; only the new widget created by my grid component would get its enclosing React context.
> 
> In general the two-way binding nature of GridStack and the one-way dataflow philosophy of React are in conflict, and the compatibility is more tenuous than the docs let on. I wish GridStack had a mode where it gives me change events without doing any DOM manipulation at all, and leaves it up to me to accomplish the DOM manipulations.

Faced the same problem. To synchronize data, there must be a single entry and exit point, in my case for vue, this is a variable in the v-model. I don't need to control the DOM via gridstack, but it's enough to make changes in the v-model. I would like to implement a mode for the grid that would disable all dom manipulations for drag and drop events, resizing, and so on. Does anyone have any idea how to do this?

adumesny:
so I use gridstack at work (or used to) with an angular wrapper that has a 'fd-grid' and 'fd-griditem` component that uses Angular dynamic object creation to create the Angular based layout dynamically from some REST json. during drag from outside (toolbar of widgets) I use helper placeholder images and on drop I replace them with the real Angular component (that gets created on the fly, just like during de-serialization). Never had an issue with dom manipulation and needing events only...

yeah there are many ways to do Angular/React/Vue wrappers. it's outside of this lib to include those wrapper (and I can't open source my version anyway) but it would be nice to have more example that do work out of the box for sure...

PR #1616, opened 2021-02-12
Add resize event listener

MrCorba:
### Description
Added an on resize listener to get the size of the node while resizing

### Checklist
- [] Created tests which fail without the change (if possible)

### DIFF
--- a/demo/events.js
+++ b/demo/events.js
@@ -19,6 +19,13 @@ function addEvents(grid, id) {
     console.log(g + 'dragstart ' + el.textContent + ' pos: (' + node.x + ',' + node.y + ') vs (' + x + ',' + y + ')');
   });
 
+  grid.on('drag', function(event, el) {
+    let node = el.gridstackNode;
+    let x = el.getAttribute('gs-x');
+    let y= el.getAttribute('gs-y');
+    console.log(g + 'drag ' + el.textContent + ' pos: (' + node.x + ',' + node.y + ') vs (' + x + ',' + y + ')');
+  });
+
   grid.on('dragstop', function(event, el) {
     let node = el.gridstackNode;
     let x = el.getAttribute('gs-x');
@@ -44,6 +51,11 @@ function addEvents(grid, id) {
     let w = el.getAttribute('gs-w');
     let h = el.getAttribute('gs-h');
     console.log(g + 'resizestart ' + el.textContent + ' size: (' + w + ' x ' + h + ')');
+  });
+
+  grid.on('resize', function(event, el) {
+    let node = el.gridstackNode;
+    console.log(g + 'resize ' + el.textContent + ' size: (' + node.w + ' x ' + node.h + ')');
   });
 
   grid.on('resizestop', function(event, el) {
--- a/doc/README.md
+++ b/doc/README.md
@@ -14,11 +14,13 @@ gridstack.js API
   - [change(event, items)](#changeevent-items)
   - [disable(event)](#disableevent)
   - [dragstart(event, el)](#dragstartevent-el)
+  - [drag(event, el)](#dragevent-el)
   - [dragstop(event, el)](#dragstopevent-el)
   - [dropped(event, previousWidget, newWidget)](#droppedevent-previouswidget-newwidget)
   - [enable(event)](#enableevent)
   - [removed(event, items)](#removedevent-items)
   - [resizestart(event, el)](#resizestartevent-el)
+  - [resize(event, el)](#resizeevent-el)
   - [resizestop(event, el)](#resizestopevent-el)
 - [API](#api)
   - [addGrid(parent: HTMLElement, opt: GridStackOptions)](#addgridparent-htmlelement-opt-gridstackoptions)
@@ -189,6 +191,15 @@ grid.on('dragstart', function(event: Event, el: GridItemHTMLElement) {
 });
 ```
 
+### drag(event, el)
+
+called when grid item is dragged
+
+```js
+grid.on('dragstart', function(event: Event, el: GridItemHTMLElement) {
+});
+```
+
 ### dragstop(event, el)
 called after the user is done moving the item, with updated DOM attributes.
 
@@ -235,6 +246,15 @@ called before the user starts resizing an item
 
 ```js
 grid.on('resizestart', function(event: Event, el: GridItemHTMLElement) {
+});
+```
+
+### resize(event, el)
+
+called while the user is resizing an item
+
+```js
+grid.on('resize', function(event: Event, el: GridItemHTMLElement) {
 });
 ```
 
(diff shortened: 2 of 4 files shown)


demo/events.js
```diff
@@ -46,6 +46,10 @@ function addEvents(grid, id) {
     console.log(g + 'resizestart ' + el.textContent + ' size: (' + w + ' x ' + h + ')');
   });
 
+  grid.on('resize', function(event, newWidget) {
```
adumesny, line 49: make it el to be consistent with others and pass the DOM element, not the node...

doc/README.md
```diff
@@ -238,6 +239,15 @@ grid.on('resizestart', function(event: Event, el: GridItemHTMLElement) {
 });
 ```
 
+### resize(event, el)
+
+called while the user is resizing an item, with current node attributes
+
+```js
+grid.on('resize', function(event: Event, item: GridStackNode) {
```
adumesny, line 247: don't pass the node, pass the element (from which node can be accessed) to be consistent... then they can use the actual px width/height or node x,y (column based)

src/gridstack-dd.ts
```diff
@@ -449,6 +449,15 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
     this.engine.moveNode(node, x, y, w, h);
     if (resizing && node.subGrid) { (node.subGrid as GridStack).onParentResize(); }
     this._updateContainerHeight();
+
+    if (this._gsEventHandler[event.type]) {
```
adumesny, line 453: looks like we should also add 'drag' event... also we need to consider if we want to call it for pixel changes, or just when column/row changes happen (what you have)

Right now you have it if only column/row changes and passing node copy (why clone the values ? just pass the node as it has already moved to new location...
I would pass node.el instead to be consistent with start/top events AND add drag event as well.

src/gridstack.ts
```diff
@@ -21,7 +21,7 @@ export interface GridHTMLElement extends HTMLElement {
   gridstack?: GridStack; // grid's parent DOM element points back to grid class
 }
 export type GridStackEvent = 'added' | 'change' | 'disable' | 'dragstart' | 'dragstop' | 'dropped' |
-  'enable' | 'removed' | 'resizestart' | 'resizestop';
+  'enable' | 'removed' | 'resizestart' | 'resizestop' | 'resize';
```
adumesny, line 24: alphabetize please

adumesny (changes requested):
thank you for making this suggestion! Please make the corrections listed.

src/gridstack-dd.ts
MrCorba: I had a problem getting the Element with the updated values. I've overlooked the node.el, this does the trick.

I added it to column/row changes, because I think it makes the most sense regarding snapping. The moment it snaps to another grid you want an update and do something with that

MrCorba: I've changed it to use the target, but I had to use ``_writePosAttr`` to get the current attributes while resizing

demo/events.js
```diff
@@ -19,6 +19,13 @@ function addEvents(grid, id) {
     console.log(g + 'dragstart ' + el.textContent + ' pos: (' + node.x + ',' + node.y + ') vs (' + x + ',' + y + ')');
   });
 
+  grid.on('drag', function(event, el) {
+    let node = el.gridstackNode;
+    let x = el.getAttribute('gs-x');
```
adumesny, line 24: actually I don't know why I have el.getAttribute('gs-x') vs node.x as they should be the same....maybe for debugging. I should make that clearer...

demo/events.js
```diff
@@ -46,6 +53,12 @@ function addEvents(grid, id) {
     console.log(g + 'resizestart ' + el.textContent + ' size: (' + w + ' x ' + h + ')');
   });
 
+  grid.on('resize', function(event, el) {
+    let w = el.getAttribute('gs-w');
```
adumesny, line 57: add let node = el.gridstackNode; (which is simpler to use really)

doc/README.md
```diff
@@ -238,6 +249,15 @@ grid.on('resizestart', function(event: Event, el: GridItemHTMLElement) {
 });
 ```
 
+### resize(event, el)
+
+called while the user is resizing an item, with current node attributes
```
adumesny, line 254: remove "with current node attributes"

src/gridstack-dd.ts
```diff
@@ -449,6 +449,13 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
     this.engine.moveNode(node, x, y, w, h);
     if (resizing && node.subGrid) { (node.subGrid as GridStack).onParentResize(); }
     this._updateContainerHeight();
+
+    if (this._gsEventHandler[event.type]) {
+      let target: GridItemHTMLElement = event.target as GridItemHTMLElement;
```
adumesny, line 454: is target != node.el ?
I need to think about are we calling about placeholder (which is grid constrained) or the actual content is that is freely being resized by the mouse (pixel following)... I would have to debug it and see what makes sense. if you are doing the live item when it won't match the mouse...

Also I would not conditionally call _writePosAttr() if you have an event as it's changes the behavior and hard to debug/test. Always do this (assuming we need that - which I don't believe we do on the placeholder as it need to reflect the new column/row sizing.

adumesny (changes requested):
thanks for the change. 
trying to understand what you need "live" resizing events because if you are updating the content inside, the live picture is supposed to follow the mouse cursor NOT the grid snapping dimensions (which is what the placeholder does but is empty). I'm wondering if we asking for problem doing that....

src/gridstack-dd.ts
MrCorba: Target == node.el, but from node.el you can't go back to el.gridstackNode. Probably because it's a recursive reference.

MrCorba:
The problem I'm trying to solve with the live resizing events is that I want to verify the new widget size and do something accordingly. Let's say there is a widget with a chart that is 4x4.  When the tile gets smaller than 4x2, the chart doesn't fit anymore. So I would like to display a message/give a red border from a certain size, to warn the user that the size is too small and the chart will be removed (or something like that)

I know I can use min width/height to force minimal sizes, but I want the user to be able to make it smaller and deal with the consequences.

src/gridstack-dd.ts
adumesny: node.el.gridstackNode should work - I should always have that circular link, but placeholder vs real widget being dragged/resized might be different. Again I'll have to look.

src/gridstack-dd.ts
```diff
@@ -452,6 +453,11 @@ GridStack.prototype._prepareDragDropByNode = function(node: GridStackNode): Grid
       if (resizing && node.subGrid) { (node.subGrid as GridStack).onParentResize(); }
       this._updateContainerHeight();
     }
+    this._writePosAttr(target, node.x, node.y, node.w, node.h);
```
adumesny, line 456: this all needs to move inside the if (this.engine.moveNodeCheck(node, x, y, w, h)) {}
otherwise it didn't change (constrain).

adumesny:
thanks, I will tweak it further.

PR #1617, opened 2021-02-12
Add array sort compareFunction to keep items respective DOM order in Firefox

marcel-necker:
### Description
grid-stack-items does not keep their respective DOM order if x,y declarations are missing (autoPosition) in Firefox.
The array sort order is implementation-defined if their is not a consistent comparison function for the elements of items:
https://tc39.es/ecma262/#sec-array.prototype.sort

Following simple example shows an reversed order of the grid-stack-items in Firefox than they appear in the DOM:
https://jsbin.com/yitiputuqe/1/edit?html,output

### Checklist


adumesny:
thank you for the fix and example! didn't know FF needed a compare... I'll update the readme.

PR #1619, opened 2021-02-12
fix event handler callback signature

wurambo:
### Description
Creates a shared GridStackEventHandlerCallback type and updates the event handler callback signature to account for optional helper parameter

### Checklist
- [] Extended the README / documentation, if necessary

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -8,7 +8,7 @@
 
 import { GridStackEngine } from './gridstack-engine';
 import { obsoleteOpts, obsoleteAttr, Utils, HeightData } from './utils';
-import { GridStackElement, GridItemHTMLElement, GridStackWidget, GridStackNode, GridStackOptions, numberOrString, ColumnOptions } from './types';
+import { ColumnOptions, GridItemHTMLElement, GridStackElement, GridStackEventHandlerCallback, GridStackNode, GridStackOptions, GridStackWidget, numberOrString } from './types';
 import { GridStackDDI } from './gridstack-ddi';
 
 // export all dependent file as well to make it easier for users to just import the main file
@@ -828,8 +828,7 @@ export class GridStack {
    * grid.el.addEventListener('added', function(event) { log('added ', event.detail)} );
    *
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public on(name: GridStackEvent, callback: (event: Event, arg2?: GridItemHTMLElement | GridStackNode[]) => void): GridStack {
+  public on(name: GridStackEvent, callback: GridStackEventHandlerCallback): GridStack {
     // check for array of names being passed instead
     if (name.indexOf(' ') !== -1) {
       let names = name.split(' ') as GridStackEvent[];
--- a/src/types.ts
+++ b/src/types.ts
@@ -26,6 +26,8 @@ export interface GridItemHTMLElement extends HTMLElement {
 
 export type GridStackElement = string | HTMLElement | GridItemHTMLElement;
 
+export type GridStackEventHandlerCallback = (event: Event, arg2?: GridItemHTMLElement | GridStackNode | GridStackNode[], newNode?: GridStackNode) => void;
+
 /**
  * Defines the options for a Grid
  */


src/gridstack-dd.ts
```diff
@@ -25,7 +25,7 @@ export type DDKey = 'minWidth' | 'minHeight' | 'maxWidth' | 'maxHeight';
 export type DDValue = number | string;
 
 /** drag&drop events callbacks */
-export type DDCallback = (event: Event, arg2: GridItemHTMLElement, helper?: GridItemHTMLElement) => void;
+export type DDCallback = GridStackEventHandlerCallback;
```
adumesny, line 28: no point in defining a type from another type!

src/gridstack-dd.ts
```diff
@@ -9,7 +9,7 @@
 
 import { GridStackDDI } from './gridstack-ddi';
 import { GridItemHTMLElement, GridStackNode, GridStackElement, DDUIData, DDDragInOpt } from './types';
-import { GridStack } from './gridstack';
+import { GridStack, GridStackEventHandlerCallback } from './gridstack';
```
adumesny, line 12: you have it defined on type, not gridstack per say....

src/gridstack.ts
```diff
@@ -828,8 +837,7 @@ export class GridStack {
    * grid.el.addEventListener('added', function(event) { log('added ', event.detail)} );
    *
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public on(name: GridStackEvent, callback: (event: Event, arg2?: GridItemHTMLElement | GridStackNode[]) => void): GridStack {
+  public on(name: GridStackEvent, callback: GridStackEventHandlerCallback): GridStack {
```
adumesny, line 840: not correct because you changed arg2 from GridItemHTMLElement | GridStackNode[] to only GridItemHTMLElement 
we use the array of nodes in change and other callbacks.

adumesny (changes requested):
not sure what that buys you as you didn't change any code per say.... it's also incorrect.

wurambo:
@adumesny Apologies, I was merely trying to fix the issue of a mismatched type.  To account for the dropped event listener `arg2` needs to be updated to `GridItemHTMLElement | GridStackNode | GridStackNode[]` and a third optional `helper` of type `GridStackNode` parameter is also needed

snippet below causing typescript error
```
newGrid.on('dropped', onDroppedCallback);

Argument of type '(event: Event, previousWidget: GridStackNode, newWidget: GridStackNode) => void' is not assignable to parameter of type '(event: Event, arg2?: GridStackNode[] | GridItemHTMLElement | undefined) => void'.
```

src/gridstack-dd.ts
wurambo: Does DDCallback need to be its own type? It should be the same type as GridStackEventHandlerCallback so I was hoping we could share the types

src/gridstack-dd.ts
wurambo: this should be from types, my VScode automatically imported it from the wrong place. Will keep note 👍

src/gridstack.ts
wurambo: Yep, you are correct. Apologies I just started working with this library and trying to help contribute a small fix. `arg2` can also be GridStackNode for the dropped event listener and also we need to account for a third `helper` param of type `GridStackNode`

src/gridstack-dd.ts
adumesny: not if it's the same, but it is different.

src/gridstack-dd.ts
adumesny: yeah VS code that sometimes...

adumesny:
yes you are right! dropped event sends 2 nodes... didn't catch that. `DDCallback ` should not change, but `GridStackEventHandlerCallback` shoudl have those 2 changes instead.

wurambo:
Thanks for the help and also the quick response! I've updated the PR. If you'd like any other changes please let me know and I can take care of it

src/gridstack.ts
```diff
@@ -8,7 +8,16 @@
 
 import { GridStackEngine } from './gridstack-engine';
 import { obsoleteOpts, obsoleteAttr, Utils, HeightData } from './utils';
-import { GridStackElement, GridItemHTMLElement, GridStackWidget, GridStackNode, GridStackOptions, numberOrString, ColumnOptions } from './types';
+import {
```
adumesny, line 11: revert to be consistent please.

src/types.ts
```diff
@@ -26,6 +26,8 @@ export interface GridItemHTMLElement extends HTMLElement {
 
 export type GridStackElement = string | HTMLElement | GridItemHTMLElement;
 
+export type GridStackEventHandlerCallback = (event: Event, arg2?: GridItemHTMLElement | GridStackNode | GridStackNode[], helper?: GridStackNode) => void;
```
adumesny, line 29: let's call 3rd one newNode (I believe only used for 'dropped' case

adumesny (changes requested):

wurambo:
Updated! Do you need me to squash my commits down or do you squash merge?

adumesny:
yes please, squash to I can review a final one (I can squash during commit but not to view final)

src/gridstack.ts
```diff
@@ -8,7 +8,7 @@
 
 import { GridStackEngine } from './gridstack-engine';
 import { obsoleteOpts, obsoleteAttr, Utils, HeightData } from './utils';
-import { GridStackElement, GridItemHTMLElement, GridStackWidget, GridStackNode, GridStackOptions, numberOrString, ColumnOptions } from './types';
+import { ColumnOptions, GridItemHTMLElement, GridStackElement, GridStackEventHandlerCallback, GridStackNode, GridStackOptions, GridStackWidget, numberOrString } from './types';
```
wurambo, line 11: do you want it reverted back to unalphabetical order? Would be nice to introduce a code formatter like prettier

adumesny: alpha is best. keep it. thanks.

wurambo:
👍 Updated for final review

Issue #1605, opened 2021-02-01
Its not possible to move item A at bottom of item B


zgilewski:
We are not able to correctly move one element below another one.

gridstack version: 2.1.0

![image](https://user-images.githubusercontent.com/28985410/106481851-25216f00-64ad-11eb-8450-e162bc49c270.png)

We should be able to move element 1 below element 2.
Currently its not possible. The only way is to drag element 2 above element 1.

adumesny:
dup of #1094, and you need to use latest rev before posting bugs as well. thanks.

zgilewski:
Its not about latest rev, as this is also happening in gridstack demos public available as you can see on attached screen, which I think using latest version of gridstack. And it is not duplication of 1094 as this describes different behaviour, or every problems with placing elements in the grid is treated as drag heuristic issue.

adumesny:
then you need to provide a jsfiddle or other example and not expect me to replicate this for you. There are other opene bugs about hard to place below - so you not possible. which is it. again post a specific example and I will re-open.

zgilewski:
Hmm, but I didnt write any custom code. To check if this issue exist in gridstack lib - not in my code, I just played with demo from https://gridstackjs.com/demo/float.html. In that page just float to **false**, add one more widget and try to move one below other. Nothing else is needed to reproduce this issue.

adumesny:
ha yes I can see it now, likely cause by another recent fix I made about dragging out (likely #1580).
would have been nice to see running video or steps you just gave me.

pcardinaliFTX:
https://user-images.githubusercontent.com/78551686/106914664-86437f80-6705-11eb-83f6-7083c4c60b6f.mov

here is the video as per zgilewski instructions

Issue #1094, opened 2019-12-16
drag heuristics can make it hard to place items (ex: web demo)


YuriGal:
Is there a way to alter behavior when one widget is dragged on top of other?  In the demo below I am trying to insert widget "7" between "9" and "2". If I drag it directly into the gap - it assumes I want insert "7" above "2" and pushes "2" down even thought I drag "7" thru the lower portion of "2". Only if I drag "7" all the way around and over "9" directly - it moves "9" down.

<img src="https://i.imgur.com/pxjSzX0.gif">

Is it possible to keep the widget that is being dragged over in place to a certain point (say when widget being dragged is at over 50% of its height) ?

adumesny:
I agree, the heuristics could use some tweaking, and be more like over the second half vs the first half - it's also not clear if it uses the mouse position or just the box outline (or something in between).

ataft:
I'm having a similar issue when trying to drag to the bottom. As you can see in the demo below, I have to drag the widget about 90% over the bottom widgets in order for it to move:

![gridstack_issue](https://user-images.githubusercontent.com/11670864/73802061-40aa4a80-4771-11ea-9318-8515f7c29f87.gif)

Prid13:
Any update on this? I'm looking into gridstack.js and it's quite promising, but the weird behavior when dragging and dropping is counter intuitive to me.

Like @ataft demonstrated, dragging an element **below** another requires about 90% of height to be below the target in order to work.

Any fix in the works, or options to tweak this as per personal preference? :)

adumesny:
on my very long todo list... but first get 2.0 out, then likely focus on this.

ranjeettamil:
@adumesny  Waiting for features from very long time, 
Please add this feature in upcoming release.

adumesny:
@ranjeettamil are you willing to pay for it ? there is a donate button (I sure didn't see your $$ contributions) and very few people bother contributing or helping with this project, yet expect things to be magically done for them for free... and silence...

Prid13:
@adumesny: I'm sure he didn't mean it like that. How big of a donation are we talking about to get this feature in motion asap? I'm willing to chip in and @ranjeettamil maybe you are too? :)

adumesny:
@Prid13 only @ataft has donated $50 so far which is very much appreciated. heck, if most people that use gridstack would even give 'beer money', someone could possibly make this a fulltime gig... never heard from @ranjeettamil again.
[...]

PR #1621, opened 2021-02-15
collision overall - support for swap

### LINKED ISSUE #149: Cannot move widget horizontally (Allow horizontal swapping without creating new rows)
for example : i have a grid with fixed 4 columns and four rows When i was trying to move the widget **Horizontally** and no place for the other widgets to slide up or down , the end result is you will not be able to move that widget unless there is enough space for other widgets to slide up/down.

is there any way to make the other widgets to **slide left/right**? so i can move any widget when i have a fixed width and height grid with full capacity widgets.

adumesny:
### Description
big overall of the collision detection code - you can now:
* swap items of same size vertically/horizontally when grid is full (`maxRow`)
**this has been 5.5years in the making and the most asked question**
It is now also the default in `float:false` as it feels more natural than pushing new row
(could add Alt-key to get either behavior of push vs swap ?)
* moving items down or up behave the same way (used to have to push WAY past to insert after)
* optimized the collision code to not do extra work multiple times and only check
if change and not tried before
* heuristics now checks for >50% past the grid mid point (taking margin into account)a

TODO part II:  handle mid point of dragged over items rather 50% of row/column
and check for the most covered when multiple items collide.

* fix #149 #1605 and partial #1094 (need better support for larger items mid point)

### Checklist

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -50,6 +50,9 @@ Change log
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 ## 3.3.0-dev
 
+- fix [#149](https://github.com/gridstack/gridstack.js/issues/149) [#1094](https://github.com/gridstack/gridstack.js/issues/1094) [#1605](https://github.com/gridstack/gridstack.js/issues/1605) re-write of the **collision code**! you can now swap items of the same size (vertical/horizontal) when grid is full, and is the default in `float:false` (top gravity) as it feels more natural. Could add Alt key for swap vs push behavior later.<br>
+Dragging up and down now behave the same (used to require push WAY down past to swap/append). Also much more efficient collision code.<br>
+Still TODO: handle mid point of dragged over items rather 50% of row/column and check for the most covered when multiple items collide.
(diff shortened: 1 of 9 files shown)


adumesny:
default `float:false` (top gravity)

https://user-images.githubusercontent.com/3496166/107982087-4340a200-6f78-11eb-8b39-5a9b448a610d.mp4


`float:true` case

https://user-images.githubusercontent.com/3496166/107982494-045f1c00-6f79-11eb-9d84-14876431287a.mp4


`maxRow:3` for full grid, even if float=true

https://user-images.githubusercontent.com/3496166/107982687-5c961e00-6f79-11eb-98fd-a114196b3381.mp4

harrington101:
Is there a workaround for the float true case? I do not have a max rows and would like to drag row 3 to 4 and have them swap. Instead, 4 is pushed down and an empty space is left where 3 was.

adumesny:
@harrington101 that is not the design, so no 'workaround'. if you don't want empty spaces, that is what `float:true` is for (no empty vertical spaces). If you want some hybrid, you can write your own collision with 5.x letting you specify you own engine...

harrington101:
I'm not looking for a hybrid. Instead, I'm looking to have box 1 swap with box 2 rather than running away from it. I do want empty spaces which is why I set float to true. The issue is that box 2 moves unexpectedly.

PR #1628, opened 2021-02-20
collision overall Part II - 50% coverage check

adumesny:
### Description
more fix for #1094 (part II of #1621)
* we now check for >50% coverage (using pixel drag vs target region) before
we collide during a drag to replace
* this means we now support larger size swapping at their mid point.
target is always used for mid point.

**Note**: web demo (#1094) is still not totally ideal as item 9 goes side way for some reason...

### Checklist
- [X All tests passing (`yarn test`)


adumesny:
showing 50% coverage (sized 2 instead of 1 items)
https://user-images.githubusercontent.com/3496166/108586804-95c9e780-7305-11eb-8733-be7dec94284f.mp4

web demo (still some issues left, especially placing below)
https://user-images.githubusercontent.com/3496166/108586908-0a048b00-7306-11eb-97f1-8ae7adda33fe.mp4

**this stuff is HARD**!!

PR #1630, opened 2021-02-21
collision: re-pack in middle of collision

adumesny:
### Description
in order to improve #1094 behavior, we may now do a partial repack after we moved an item below
keeping the original wanted location so we can try again
and push items out of the way.

### Checklist


adumesny:
https://user-images.githubusercontent.com/3496166/108612321-716d1a00-739c-11eb-8d71-70c8d69f7405.mp4

PR #1651, opened 2021-03-07
Adding Vue 3 Composition API

RiFrost:
### Description
To make the difference between Vue 2 and Vue 3 even clearer, perhaps using the Composition API would be a good idea. I stumbled across this because I only started learning Vue with version 3.x, so the Composition API is more familiar to me than the Vue 2 approach. I could imagine that people often feel this way and therefore maybe both approaches could be covered to save time in translation. 

Since this is my first real PR in an open source project, I'm not sure if this is a good idea and if this is the way to do it. 

I am open for criticism and feedback also regarding the PR.


demo/vue3js.html
```diff
@@ -1,88 +1,97 @@
...
-            };
-            node.id = node.content = String(this.count++);
-            this.grid.addWidget(node);
-          },
-        },
-      }).mount("#app");
-    </script>
-  </body>
-</html>
+<!DOCTYPE html>
```
adumesny, line 1: can you make the indentation (2 spaces) and line ending the same so we can see what changed ? thanks

RiFrost:
I dont think that this what i did will help you. I am sorry for that. But I think there is the Option to not show Whitespace changes when comparing files which should also resolve in seeing the actual changes better.

If this is not an Option i could make a new PR or otherwise you may help me sorting this out the right way. Sorry ^^

demo/vue3js.html
```diff
@@ -1,88 +1,97 @@
...
+      <p>
+        If your app requires more complex render logic than the inline template
+        in `addWidget`, consider
+        <a
+          href="https://github.com/gridstack/gridstack.js/tree/develop/doc#makewidgetel"
+          >makeWidget</a
+        >
+        to let Vue deal with DOM rendering.
+      </p>
+      <button type="button" @click="addNewWidget()">Add Widget</button> {{ info
```
adumesny, line 28: not a useful change. please revert

demo/vue3js.html
```diff
@@ -1,88 +1,97 @@
...
+    </main>
+    <script type="module">
+      import { createApp } from "https://cdn.jsdelivr.net/npm/vue@3.0.0/dist/vue.esm-browser.js";
+      import { ref, onMounted } from "vue";
+
+      createApp({
+        setup() {
+          let count = ref(0);
+          let info = ref("");
+          let grid = null;
```
adumesny, line 40: did you look at this and is it not an issue anymore ? great if this works as expected now.

// grid: undefined, // <-- don't do that. see https://github.com/gridstack/gridstack.js/issues/1635

demo/vue3js.html
```diff
@@ -1,88 +1,97 @@
...
+          const items = [
+            { x: 2, y: 1, h: 2 },
+            { x: 2, y: 4, w: 3 },
+            { x: 4, y: 2 },
+            { x: 3, y: 1, h: 2 },
+            { x: 0, y: 6, w: 2, h: 2 },
+          ];
+
+          onMounted(() => {
+            grid = GridStack.init({
```
adumesny, line 50: no this.grid needed ? how is that scoped then ????
(don't want it global but this Vue component...)

I would insert the comment back
// Provides access to the GridStack instance across the Vue component.

adumesny (changes requested):
yes I can merge and skip white spacing but still shows newline ending ?

Also best if you can merge your changes - git rebase -i HEAD~3  (I can squash on submit otherwise)

demo/vue3js.html
RiFrost: From my understanding you got access to the grid within the setup function at all time. So it is only accessible for the Vue component itself. 
So you are able to define functions or vars etc. inside the setup function which are able to work with the grid.

demo/vue3js.html
RiFrost: Since it is all inside the setup funtion i have not had any issues with that.

demo/vue3js.html
adumesny: ok, did you want to make line ending change and merge into 1 review, then I can accept it. thank you!

PR #1654, opened 2021-03-09
adding vue 3 composition api

RiFrost:
### Description

To make the difference between Vue 2 and Vue 3 even clearer, perhaps using the Composition API would be a good idea. I stumbled across this because I only started learning Vue with version 3.x, so the Composition API is more familiar to me than the Vue 2 approach. I could imagine that people often feel this way and therefore maybe both approaches could be covered to save time in translation.

I started a new PR because my branch got a little messy. I hope that the spacing and line ending is fine this time ^^


adumesny:
thank you!

PR #1663, opened 2021-03-20
collision: more edge case fixes

### LINKED ISSUE #1094: drag heuristics can make it hard to place items (ex: web demo)

adumesny:
### Description
fix #1094

* found I needed to take entire row (not just new location) when doing collision to correctly push down bottom items first so they keep their order and prevent large objects from leapfrogging smaller ones.
(like old code used to do)
* removed _beforeDrag and _packY fields and re-use existing _orig position data which simplifies things
* removed yOffset during _fixCollisions() as it was incorrect (reverse sorting should take care of that)
* on _leave() of external item, we now restore all items to original position using restoreInitial() (prevents any error from sticking)

NEEDLESS to say, this is MUCH more complicated than the old _fixCollision() that moved to new loc then pushed anything down in that row and below (bottom first) which is dead simple... but required you to move 90% past the item when moving down.

### Checklist

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,10 @@
-The MIT License (MIT)
+v4 Copyright (c) 2021 Alain Dumesny
+under MIT license
 
+previous v3.x and older is:
 Copyright (c) 2014-2020  Alain Dumesny, Dylan Weiss, Pavel Reznikov
+
+The MIT License (MIT)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/demo/two.html
+++ b/demo/two.html
@@ -95,9 +95,9 @@
     let grids = GridStack.initAll(options);
     grids[1].float(true);
 
-    // new 4.x static method instead of setting up options on every grid (never been per grid really but old options still work)
+    // new 4.x static method instead of setting up options on every grid (never been per grid really) but old options still works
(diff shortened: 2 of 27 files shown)


Prid13:
Thank you soo much for you hard work! I really, truly appreciate it and wish more people could donate to reward your continued effort. I'm just really glad you took the time to do this :)

adumesny:
@Prid13 thank you for the comment. Would be great if you would follow on that donation. I've been thinking of making this commercial version as it take a huge amount of time and very few do contribute.

Prid13:
I've already donated a while back but will try donating a bit more (won't be as much, unfortunately) :) I'm thinking of using this in a commercial product myself and will be sure to donate more if I ever manage to turn a profit on it :)

adumesny:
sorry, couldn't connect to name. Zain Mushtaq ?

Prid13:
Yes :) but could you please delete that? Trying not to mix my online and personal identity haha 😅

Issue #1704, opened 2021-04-01
4.x: style="position: sticky" messing up Dragging


Dodger77:
## Subject of the issue
I am trying to make a lot of custom widgets to add to multiple grids. In order to access all of the widgets these stay in a sticky positioned `<div>` (`position: fixed` behaved the same way). Regardless of using float option true/false in the second grid it is not possible to position the dragged in widget under the other widgets.

## Your environment
* Version 4.0.3 tested in HTML5 and jQuery versions
* Browser tested: Firefox (Linux), Chrome (Linux), Edge (Windows 10)

## Steps to reproduce
https://codepen.io/Dodger77/pen/KKaaaJN (HTML5 version with `float: true`)
https://codepen.io/Dodger77/pen/qBRRvEm (jQuery version with `float: false`)

## Expected behavior
Get new dragged in widgets positioned under the existing widgets.

## Actual behavior
New dragged in widgets will always be on top of the second grid when dragged in.

https://user-images.githubusercontent.com/255767/113271935-5e266600-92db-11eb-9920-a2ee06de2838.mp4

Dodger77:
Here as a comparison the same code with gridstack.js 3.3.0:

https://codepen.io/Dodger77/pen/GRrrevw

adumesny:
I see the issue https://jsfiddle.net/adumesny/esyLr4hf/
looks like ` <div style="position: sticky; top: 0;">` is messing up the scroll bar computation for the second grid.

adumesny:
fixed in next release. don't forget to donate if you find lib useful!

Dodger77:
Thank you very much!

Issue #1576, opened 2021-01-16
Dragging will endlessly scroll/too fast.


bcbuddy:
## Subject of the issue
First and most critical for me, you can drag an item infinitely down the page. 

Secondly, it seems that with the new HTML5 DD update, the "scroll" property in dragInOptions is always set as "true" and we have no control over it.

Finally, the speed in which an item drags is incredibly fast. It would be nice to have some control over it.

Issues 1 & 3 combined can be very annoying together.

## Your environment
3.1.4 with HTML5 DD

## Steps to reproduce
Use your advance demo: https://gridstackjs.com/demo/web2.html
Grab an item in the grid, drag it to the bottom of the page.
Note how fast it moves and that it will drag forever. This is a real pain when a user needs to drag something around at the bottom of a grid container.

## Expected behavior
Most importantly, I would like the infinite dragging fixed. An item should not be dragged past the bottom of the page(or it's container if specified? I know you guys noted that containment is not supported yet with HTML5 - hopefully soon!). 
I would like the option to enable/disable scrolling.
I would also like the option to increase/decrease sensitivity of the drag

## Actual behavior
Items can be dragged forever down the page.

adumesny:
dragging/resizing past the bottom is by design (see #1001) but I agree that it is way too fast... not sure we need option to turn it off if designed correctly (flag seems like a band-aid to me).

looking at code commend (which I have not verified) rhlin says:
```js
  scroll?: boolean; // nature support by HTML5 drag drop, can't be switch to off actually
```
but I'm having so many html draggable issues (on mac, safari) that I started implementing plain mousedown/mousemove instead, in which case we could possibly support whatever options we want.

but this is a all a ton of work, so help or donation would be appreciated.

adumesny:
well don't worry, in v6 it apparently doesn't scroll at all :) - see #2032

PR #1727, opened 2021-04-12
fix resize-scroll when gridster is not at top of page

Manfred-on-github:
### Description
when scroll element does not start from top of page, the scrolling during resize does not work as expected, the decision if to scroll, and amount to scroll, is computed wrong.
This PR fixes this.

### Checklist
- [x ] All tests passing (`yarn test`)

### DIFF
--- a/spec/e2e/html/1727_resize_scroll_top.html
+++ b/spec/e2e/html/1727_resize_scroll_top.html
@@ -1,0 +1,56 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="stylesheet" href="../../../demo/demo.css"/>
+  <!-- <script src="https://cdn.jsdelivr.net/npm/gridstack@3.3.0/dist/gridstack-h5.js"></script> -->
+  <script src="../../../dist/gridstack-h5.js"></script>
+  <style>
+    .topHeader {
+      background: rgb(127, 87, 180);
+      height: 200px;
+    }
+    .willScroll {
+      height: 600px;
+      overflow-y: auto;
+      flex-grow: 1;
+    }
+    .row {
+        display:flex;
+        width: 100%;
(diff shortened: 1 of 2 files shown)


adumesny:
@Manfred-on-github did you test with latest 4.2.0 as I made some changes to scrollbar during `dropover` (fix #1704) which may or may not be related... would be great to have an example showing the issue before.

thanks for doing this fix!

Manfred-on-github:
have added a test page. Have checked error can be reproduced in 4.2.0, 4.0.3 and 3.3.0.

adumesny:
thanks, that is perfect.

adumesny:
funny you called it 'gridster' :) did you used that lib before ??

Manfred-on-github:
oh - yes, we used that originally, then, looking for something that is still maintained, we found gridstack. - Thank you for keeping this up.

adumesny:
while this fixes the offset issue of the grid, still doesn't work perfectly as it scroll too soon... why does it scroll when we hit the white sidebar and not the actual end of the yellow grid ?

Manfred-on-github:
for me it scrolls when the y-position of the pointer moves outside the range indicated by the white sidebar.

This seems to be by design, as the function updateScrollResize() is invoked from gridstack-dd.ts with parameter distance as cellHeight, therefore (to visually show this in my sample) have explicitly set grid rowHeight to 130px, which is also used as border height in the sidebar.

Of course this might be changed to some smaller value. For me it is also ok as it is now, as widgets further down will appear in view early before the resize just pushes them down, so user better sees what resizing of a particular widget will mean to positions of other widgets.

Something that would rather improve user experience, is connected to the comment in that function about using a timer. As it is now, when pointer gets near the edge, this function scrolls the page, but then the pointer is still near the edge, so condition to scroll still holds, and the next scroll happens immediately, and so on. So something that limits these scroll events would definitely help here.

adumesny:
yes about timer, and what #1576 is about (at least the speed part which I agree and needs to get done).

adumesny:
@Manfred-on-github looks like your change caused this issue. I forced `offsetTop = 0` (revert your change) and back to working as expected...

Issue #1745, opened 2021-04-23
Resizing scrolling jumping around in demo page, doesn't size correctly


Helyan47:
## Subject of the issue
Resizing an element in Firefox causes an infinite scroll to occur and its size cannot be modified 

## Your environment
* Demo of the webpage https://gridstackjs.com/#getStarted
* Firefox / Windows10

## Steps to reproduce
Resize an element in Firefox

## Expected behavior
Resize shouldn't cause an infinite scroll

## Actual behavior
Resize an element cause infinite scroll in Firefox

https://user-images.githubusercontent.com/54754636/115926956-fa2f3180-a483-11eb-9b61-a0f2978d98e1.mp4

adumesny:
happen on Chrome win10 as well. The bouncing around whie scrolling and not sizing correctly issue was caused by #1727 in v4.2.1
(the infinite scrolling part is a feature)

adumesny:
fixed in next release.

PR #1751, opened 2021-05-07
fix scroll problem when scrollbar on html element

Manfred-on-github:
### Description
fix for issue #1745 

looks like browsers are quirky beyond believe.
The properties .clientHeight and .getBoundingClientRect() are inconsistent for regular elements compared to the outmost 'html' element.

Hope you don't mind the verbose comment, and sorry for being late to look at this problem.

Have also enhanced test file covering previous problem of #1727 to contain an additional scrollbar on html element to see any adverse effects, of course such scrollbar defeats the concept of automatically scrolling while user drags something, since you don't know what needs to be moved.

### Checklist
- [x] Created tests which fail without the change (this is your test file website.html)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -351,18 +351,19 @@ export class Utils {
   static updateScrollResize(event: MouseEvent, el: HTMLElement, distance: number): void {
     const scrollEl = this.getScrollElement(el);
     const height = scrollEl.clientHeight;
-    //when does pointer move out of visible area, which is then fixed by scrolling that element?
-    //property clientHeight returns height of element, but for 'html', it returns
-    //height of viewport. getBoundingClientRect() returns rect of element in client
-    //coordinates, for normal elements its height equals clientHeight, but for 'html'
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -351,7 +351,18 @@ export class Utils {
   static updateScrollResize(event: MouseEvent, el: HTMLElement, distance: number): void {
     const scrollEl = this.getScrollParent(el);
     const height = scrollEl.clientHeight;
-    const offsetTop = scrollEl.getBoundingClientRect().top;
+    //when does pointer move out of visible area, which is then fixed by scrolling that element?
```
adumesny, line 354: not sure that first line help - also wrong syntax.
Also please change comments to '// ' (add space)

can we maybe rephrase this comment ? It's not helping me. I would document why we need offsetTop at all, THEN explain that document HTML is different due to returning viewport and not actual needed height. or something like that...

I have not looked into this so taking you word for it. Also add #1745 #1727 to your comment.

adumesny:
thank you!

PR #1783, opened 2021-06-11
Add keepRatio options for GridStackOptions

KadDarem:
### Description
I added a new option for Gridstack to allow keep ratio between height and width of cell

### Checklist
- [ x] Created tests which fail without the change (if possible)
- [ x] All tests passing (`yarn test`)
- [ x] Extended the README / documentation, if necessary

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -103,6 +103,7 @@ gridstack.js API
 - `handle` - draggable handle selector (default: `'.grid-stack-item-content'`)
 - `handleClass` - draggable handle class (e.g. `'grid-stack-item-content'`). If set `handle` is ignored (default: `null`)
 - `itemClass` - widget class (default: `'grid-stack-item'`)
+- `keepRatio` - if `true` cells will keep the ratio 1:1 between height and width (default: `false`)
 - `margin` - gap size around grid item and content (default: `10`). Can be:
   * an integer (px)
   * a string (ex: '2em', '20px', '2rem')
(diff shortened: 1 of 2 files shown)


adumesny:
| keepRatio` - if `true` cells will keep the ratio 1:1 between height and width (default: `false`)

if you are trying to make the cell square, and stay that way through resizing, there is already support for that with `cellHeight = 'auto'`. Your doc imply square cells but it's not, right ?

if you are trying to keep the grid itself square, then first off this var is named wrong, but I think you are trying to keep the grid itself keep the same width/height ratio even as new items are added to the bottom (taller will make it wider).

What is the use case for this and can you show a demo of the behavior ? thanks.

KadDarem:
I'm trying to keep cell square but with the possibility to define the size of cells without fixed size of the grid.
With my option, I can define the size of cell with cellHeight argument and they will keep this size even if I add column or row.

adumesny:
if you are trying to keep cell square AND define a height, you don't need this. just have `cellHeight=70` (for example) and set the grid-stack div width = 12 * 70. we only change the height as row get added/removed, we don't muck with width.

KadDarem:
It's the same for the height, I can define it in CSS but I can also use cellHeight.
But no problem, you can revoke my push. I will keep the option for my work.

adumesny:
| It's the same for the height, I can define it in CSS but I can also use cellHeight.

no it's not. gridstack will update the CSS for grid height dynamically as new rows get added/removed. that's why you have cellHeight param, to control that behavior (value 0 means you set it explicitly as well). I don't think your change is valid as you can do that today with what I mentioned above. Closing...

PR #1791, opened 2021-06-17
quickly set external helper initial position + fix scrollbar issue

nelsieborja:
### Description
This change will immediately set the initial position of the external helper, it will fix:
* issue shown in [this video](https://www.loom.com/share/904ac32dd4b44c37a60485567e85cf5e)
* by setting the correct initial `y` position of the helper, could skip unnecessary looping in `_packNodes()`

The added "fixed" position will fix the scrollbar issue when the body/container height is set to 100% (relative to window height).


adumesny:
thank you for making these changes! I had noticed the flicker as well.

Issue #1835, opened 2021-08-18
Layout incorrectly restored when node has a minimum width and layout changes from 12 -> 1 -> 12


hssm:
Nodes with a minimum width that is larger than the transitioning column count (e.g: minW=2 when moving to 1-column mode) do not survive a layout change when the column count is restored.

I'll have a pull request with a fix shortly :)

[Reproducible example](http://jsfiddle.net/L08bq76e/2)

## Steps to reproduce

1 - Start with a large window so the grid begins in 12-column mode
2 - Reduce the window size until it resizes into 1-column mode
3 - Expand the window to return to 12-column mode

## Expected behavior
The original 12-column layout is restored.

## Actual behavior
The widgets appear to be stuck in the 1-column placement with the wrong width.

## Your environment
I've observed the issue on version 4.2.6 and 4.0.2.

adumesny:
fix in next release!

PR #1836, opened 2021-08-18
Don't override node width when switching column mode

hssm:
Provides a fix for https://github.com/gridstack/gridstack.js/issues/1835

Explanation:
When going from 12 -> 1 column, this function is run after the node has already had its width set to 1 (I will admit I don't understand the order of events perfectly, this is just what I observed). So when `node.minW` has a value, this condition overrides the node width to be larger than the column count. 

The behavior in #1835 is from a few lines down, where now the condition for `node.w > this.colum`  passes and causes the layout to be cached which was only intended to run when starting from 1 and going to 12.

### DIFF
--- a/src/gridstack-engine.ts
+++ b/src/gridstack-engine.ts
@@ -345,7 +345,7 @@ export class GridStackEngine {
 
     if (node.maxW) { node.w = Math.min(node.w, node.maxW); }
     if (node.maxH) { node.h = Math.min(node.h, node.maxH); }
-    if (node.minW) { node.w = Math.max(node.w, node.minW); }
+    if (node.minW && node.minW < this.column) { node.w = Math.max(node.w, node.minW); }
     if (node.minH) { node.h = Math.max(node.h, node.minH); }
 
     if (node.w > this.column) {


adumesny:
thanks, I'll take a look this weekend...

gb92:
Any updates on the status of this? Any way to help get it merged?

adumesny:
I need to take some time this weekend as I have a couple fixes ready to go in and release...

src/gridstack-engine.ts
```diff
@@ -345,7 +345,7 @@ export class GridStackEngine {
 
     if (node.maxW) { node.w = Math.min(node.w, node.maxW); }
     if (node.maxH) { node.h = Math.min(node.h, node.maxH); }
-    if (node.minW) { node.w = Math.max(node.w, node.minW); }
+    if (node.minW && node.minW < this.column) { node.w = Math.max(node.w, node.minW); }
```
adumesny, line 348: I think we want node.minW <= this.column as minWidth:12 should force to take 12 column for example.

Issue #1817, opened 2021-07-21
original DragEvent not passed to gridstack dropped event (no access to dataTransfer property)


zgilewski:
In griddstack 'dragged' event, we dont have access to original DragEvent, which making impossible to access dataTransfer property of DragEvent.

Gridstack version 4.0.1

We expecting that in 'dragged' event we can access original DragEvent from which we can access dataTransfer property, which will allow us to update widget content basing on data taken from dataTransfer property.
Actually its not possible.

adumesny:
can this be passed and be the same API for H5 and Jquery ? (raw event or something as they are different). Please make a code suggestion if you can, or donate for me to take a look... thanks.

zgilewski:
Problem is with H5. For Jquery it looks ok.
In case of H5 issue is in line https://github.com/gridstack/gridstack.js/blob/72d8239337b3ca43bebe1528e473d71f39ea2f93/src/gridstack-dd.ts#L280
where to gridstack dropped event handler you passing new Event object which contains ONLY type property.
But it should contains all info from original drag event, as in other case we loosing all informations.

So simple solution is to replace this line:
line 280: `this._gsEventHandler['dropped']({type: 'dropped'}, origNode && origNode.grid ? origNode : undefined, node);`
to this line:
line 280: **`this._gsEventHandler['dropped']({...event, type: 'dropped'}, origNode && origNode.grid ? origNode : undefined, node);`**

Example of usage after this change:
1. Create any html element and setup DDElement for it:
`const test = document.createElement('div');
test.className = 'xxx';
const testContent = document.createElement('div');
testContent.className = 'xxxContent';
testContent.innerText = 'ANY WIDGET';
test.appendChild(testContent);`

                const _ddElement = DDElement.init(test);
                _ddElement.setupDraggable({
                    handle: '.xxx',
                    revert: 'invalid',
                    scroll: true,
                    appendTo: 'body',
                    helper: 'clone',
                    start: (event: DragEvent) => {
                        if (event.dataTransfer) {
                            event.dataTransfer.setData('text/plain', 'BINGO');
                        }
                    },
                });`
2. for your gridstack add dropped event handler:
`grid.on('dropped', this.gridStackDropped.bind(this));
gridStackDropped(event: Event, previousWidget: GridStackNode, newWidget: GridStackNode): void {
              const dragEvent = event as DragEvent;
            if (dragEvent.dataTransfer) {
[...]

PR #1841, opened 2021-09-03
Enable passing of DragEvent to gridstack dropped event

onepartsam:
### Description
Enhancement done to allow the passing of DragEvent to gridstack dropped event, allowing access to dataTransfer property. A demo in Angular is included to showcase this enhancement.

This enhancement is related to issue #1817


demo/drag-drop-dataTransfer.ts
```diff
@@ -0,0 +1,65 @@
+/**
```
adumesny, line 1: thanks for putting an example, but user can't simply run this in a browser to test/play with it. do you mind putting a generic JS examples like the others ?
(I happen to use Angular myself but most don't as this lib isn't tied to any framework). thanks!

onepartsam: Hi, I have just included a generic JS example.

adumesny:
thank you!

onepartsam:
Hi @adumesny, thank you. May I know when you are going to make a new release with this modification?

adumesny:
I like to have a few fixes in per release if possible... I have another one pending, so maybe in a couple weeks ?

PR #1887, opened 2021-11-25
feature: support ie browser

SmileLifeIven:
### Description

### Checklist


demo/web1.html
```diff
@@ -9,10 +9,10 @@
 
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css">
   <link rel="stylesheet" href="demo.css" />
-
+  
   <script type="module" src="https://unpkg.com/ionicons@4.5.10-0/dist/ionicons/ionicons.esm.js"></script>
   <script nomodule="" src="https://unpkg.com/ionicons@4.5.10-0/dist/ionicons/ionicons.js"></script>
-
+  <script src="../src/gridstack-poly.js"></script>
```
adumesny, line 15: should be ../dist as poly should be copies to the build folder as well (or needs to)

package.json
```diff
@@ -55,15 +55,20 @@
...
     "@types/jquery": "^3.5.1",
     "@types/jqueryui": "^1.12.13",
     "@typescript-eslint/eslint-plugin": "^4.6.0",
     "@typescript-eslint/parser": "^4.6.0",
+    "babel-loader": "^8.2.3",
     "connect": "^3.7.0",
     "core-js": "^3.6.4",
-    "coveralls": "^3.0.9",
-    "doctoc": "^1.4.0",
+    "coveralls": "^2.11.4",
```
adumesny, line 70: why did you use an older one ?

package.json
```diff
@@ -55,15 +55,20 @@
   },
   "homepage": "http://gridstack.github.io/gridstack.js/",
   "devDependencies": {
+    "@babel/plugin-proposal-class-properties": "^7.16.0",
```
adumesny, line 58: remove all babel please.

package.json
```diff
@@ -77,18 +82,21 @@
...
     "serve-static": "^1.14.1",
-    "ts-loader": "^8.0.7",
+    "ts-loader": "^8.1.0",
     "typescript": "^3.7",
-    "webpack": "^5.3.2",
+    "webpack": "^5.31.2",
     "webpack-cli": "^4.6.0"
+  },
+  "dependencies": {
+    "yarn": "^1.22.17"
```
adumesny, line 100: should be a devDependencies, not dependency (that user need to install)

also yarn.lock should be updated instead... not sure why you had (?) to change the rev of the libs...

src/gridstack.ts
```diff
@@ -5,7 +5,7 @@
  * Copyright (c) 2021 Alain Dumesny
  * see root license https://github.com/gridstack/gridstack.js/tree/master/LICENSE
  */
-
+require('@babel/polyfill')
```
adumesny, line 8: no, I don't want modern browsers to compile poyfill. Only the few IE users can include our polyfill js file instead...

src/utils.ts
```diff
@@ -295,7 +295,7 @@ export class Utils {
 
   /** @internal returns the passed element if scrollable, else the closest parent that will, up to the entire document scrolling element */
   static getScrollElement(el?: HTMLElement): HTMLElement {
-    if (!el) return document.scrollingElement as HTMLElement;
+    if (!el) return document.documentElement as HTMLElement;
```
adumesny, line 298: need to understand that change as I don't want break other browser for IE, which has been dead for a while really.. so not interested in supporting it unless it's 'free'... 

maybe do || if scrollingElement doesn't exist on IE ?

webpack.config.js
```diff
@@ -7,13 +7,26 @@ module.exports = {
     'gridstack-static': './src/gridstack-static.ts'
   },
   mode: 'production', // production vs development
-  devtool: 'source-map',
-  // devtool: 'eval-source-map', // for best (large .js) debugging. see https://survivejs.com/webpack/building/source-maps/
+  // devtool: 'source-map',
+  devtool: 'eval-source-map', // for best (large .js) debugging. see https://survivejs.com/webpack/building/source-maps/
```
adumesny, line 11: can't check that in!

webpack.config.js
```diff
@@ -7,13 +7,26 @@ module.exports = {
...
+  target: ['web', 'es5'],
   module: {
     rules: [
       {
         test: /\.ts$/,
-        use: 'ts-loader',
+        use: [
+          {
+            loader: 'babel-loader',
```
adumesny, line 19: no babel please, already have webpack and ts compiler... don't think we need that many packages!

adumesny (changes requested):
thanks for making it work, although I had no interest in supporting a long dead and no longer supported browser... ok if it comes free and only IE user need to include polyfill.

demo/web1.html
SmileLifeIven: ok

SmileLifeIven:
Sorry for your time, Those are my mistakes.

src/gridstack-poly.js
```diff
@@ -202,4 +219,46 @@ if (!Array.prototype['forEach']) {
     }
     // 8. return undefined
   };
-}
\ No newline at end of file
+}
+
+// https://developer.mozilla.org/zh-CN/docs/Web/API/CustomEvent/CustomEvent
+(function () {
+  if (window.navigator.userAgent.indexOf('Chrome') > -1 ) {
```
adumesny, line 226: did you actually get that from core-js polyfill ? checking for specific browser type seems bridle to me...

src/gridstack-poly.js
```diff
@@ -146,6 +146,23 @@ if (!Array.prototype.findIndex) {
   });
 })([Element.prototype, Document.prototype, DocumentFragment.prototype]);
 
+// from: https://github.com/jserz/js_piece/blob/master/DOM/Element/prepend()/prepend().md
```
adumesny, line 149: can't access that URL... looks like a private repo. isn't there a developer.mozilla.org version instead ?

src/gridstack-poly.js
```diff
@@ -202,4 +219,46 @@ if (!Array.prototype['forEach']) {
     }
     // 8. return undefined
   };
-}
\ No newline at end of file
+}
+
+// https://developer.mozilla.org/zh-CN/docs/Web/API/CustomEvent/CustomEvent
```
adumesny, line 224: small request, can you use en-US in urls instead ? more people read english doc. thank you.

src/gridstack-poly.js
```diff
@@ -202,4 +219,46 @@ if (!Array.prototype['forEach']) {
...
+      evt.initCustomEvent( event, params.bubbles, params.cancelable, params.detail );
+      return evt;
+  }
+
+  CustomEvent.prototype = window.Event.prototype;
+
+  window.CustomEvent = CustomEvent;
+})();
+
+//https://github.com/jserz/js_piece/blob/master/DOM/ChildNode/remove()/remove().md
```
adumesny, line 242: same, can't access URL

src/gridstack-poly.js
```diff
@@ -202,4 +219,46 @@ if (!Array.prototype['forEach']) {
...
+     value: function remove() {
+       this.parentNode.removeChild(this);
+     }
+   });
+ });
+})([Element.prototype, CharacterData.prototype, DocumentType.prototype]);
+
+// suport IE
+(function () {
+  if (!document.scrollingElement) {
```
adumesny, line 261: interesting way to support field name change... is there a problem caching this though ? I'm ok with caller doing 
document.scrollingElement || document.documentElement
instead.

src/gridstack.ts
```diff
@@ -164,7 +163,7 @@ export class GridStack {
     // create the grid element, but check if the passed 'parent' already has grid styling and should be used instead
     let el = parent;
     if (!parent.classList.contains('grid-stack')) {
-      let doc = document.implementation.createHTMLDocument();
+      let doc = document.implementation.createHTMLDocument("");
```
adumesny, line 166: use '' (2 single quote) instead, also you didn't answer me if IE require a param to be passed (if so add comment)

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "outDir": "./dist",
     "sourceMap": true,
     "strict": false,
-    "target": "ES6"
+    // "target": "ES6",
+    "target": "ES5"
```
adumesny, line 19: what is the size difference between ES5 and 6 (which I believe was smaller) ? is that required for IE. need to trade off all browser vs obsolete IE

webpack.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   mode: 'production', // production vs development
   devtool: 'source-map',
   // devtool: 'eval-source-map', // for best (large .js) debugging. see https://survivejs.com/webpack/building/source-maps/
+  target: ['web', 'es5'],
```
adumesny, line 12: what does that do ?

adumesny (changes requested):
thanks, looks much better. couple comments

src/gridstack.ts
adumesny: https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument
title  Optional (except in IE)
I see that now....

adumesny:
you still have not answer my question on the size difference between ES5 and ES6 I had, and what the web target is for... and please add URL to polyfill code you add so we know where it is coming from...
looks good otherwise.

SmileLifeIven:
> you still have not answer my question on the size difference between ES5 and ES6 I had, and what the web target is for... and please add URL to polyfill code you add so we know where it is coming from... looks good otherwise.

Size:
  ES5:  gridstack-jq.js -> 202KB   gridstack-h5.js -> 81KB  gridstack-static.js -> 1KB
  ES6:  gridstack-jq.js -> 196KB   gridstack-h5.js -> 71KB  gridstack-static.js -> 40KB
    
 Web target is for some old system with original browser, they want to use new feature with they system, and there is only IE browser when some customer install system, we shoule support the feature instead of updating customer's browser.
    
And  the polyfill code source url I will add it.

SmileLifeIven:
Or we could add new script to build for es5 and relative config add into new filefolder.

adumesny:
| Or we could add new script to build for es5 and relative config add into new filefolder.

Yes, I think that would be better, because 10k for gridstack-h5.js @ 71KB is a big difference and very few if any need IE support, so I don't want to do that...

SmileLifeIven:
> | gridstack-static.js -> 1KB typo...
> 
> | Or we could add new script to build for es5 and relative config add into new filefolder.
> 
> Yes, I think that would be better, because 10k for gridstack-h5.js @ 71KB is a big difference and very few if any need IE support, so I don't want to do that. create an es5 sub-folder for it

adumesny:
re-opening in case someone else wants to continue this effort...

SmileLifeIven:
Done. Add a es5 filefolder.

es5/tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "allowSyntheticDefaultImports": true
+   */
+    "declaration": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "inlineSources": true,
+    "lib": [ "es6", "es2015", "dom" ],
+    "module": "CommonJS",
+    "noImplicitAny": false,
+    "outDir": "../dist",
```
adumesny, line 15: should be ../dist/es5 so we can have them both (I only npm publlish once)

es5/webpack.config.js
```diff
@@ -0,0 +1,22 @@
...
+          options: {
+            configFile: 'es5/tsconfig.json'
+          }
+        },
+        exclude: ['/node_modules/', '/src/h5/', '/src/index-*.ts'],
+      },
+    ],
+  },
+};
+config.output.path = path.resolve(__dirname, '../dist')
```
adumesny, line 21: dist/es5

package.json
```diff
@@ -20,6 +20,7 @@
   ],
   "scripts": {
     "build": "yarn --no-progress && rm -rf dist/* && grunt && webpack && tsc --stripInternal && yarn doc",
```
adumesny, line 22: I would have build do both es6 and es5

adumesny (changes requested):
almost there! thank you.

SmileLifeIven:
Check it.

adumesny:
thank you! I will verify it and update the documentation and publish a new version in the next couple days...

SmileLifeIven:
😊

PR #1919, opened 2022-01-12
Update gridstack.ts

ZOUJIELynn:
fix when set minRow greater than currentRow ,  the top value of placeholder is not set

### Description

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -327,7 +327,7 @@ export class GridStack {
       maxRow: this.opts.maxRow,
       onChange: (cbNodes) => {
         let maxH = 0;
-        this.engine.nodes.forEach(n => { maxH = Math.max(maxH, n.y + n.h) });
+        this.engine.nodes.forEach(n => { maxH = Math.max(maxH, n.y + n.h,this.opts.minRow) });
         cbNodes.forEach(n => {
           let el = n.el;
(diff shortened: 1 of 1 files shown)


src/gridstack.ts
```diff
@@ -327,7 +327,7 @@ export class GridStack {
       maxRow: this.opts.maxRow,
       onChange: (cbNodes) => {
         let maxH = 0;
```
adumesny, line 329: just set 
let maxH = this.opts.minRow || 0; 
instead so it doesn't check every time AND makes sure it not undefined

adumesny: code might be incorrect if undefined ? (or max(0, undefined) happens to work...

ZOUJIELynn:
> 
I set minRow =3 , but actully only two row has widgets , when i drag external widget at row 3, the top value of placeholder is not set, because there is no Css for gs-y=2

adumesny:
seems like the right fix is to make sure the dragged in item DOES increase the maxH computation instead to update the CSS rules. Separate from having minRow set as it's possible to drag an item to the bottom anyway (past the current last item).

Can you post running example by any chance ? thanks!

ZOUJIELynn:
> 

thanks , let me have a try

PR #1954, opened 2022-04-15
Create index.html

ohioh:
### Description

### Checklist


adumesny:
why would you add a top level index.html for ? also it's referring to the INSTALLED gridstack version which is NEVER the case for this source code (we generate it instead into /dist folder where the demos use.

closing.

Issue #98, opened 2015-03-22
Nested grid not retaining the relative position in vertical mode


KrishnaPG:
Thank you for sharing this good work - it is very helpful. The bootstrap style of responsiveness is very useful and the nested grids help implement very useful functions such as row or inner columns etc. (like col-xs inside col-md etc..). However, there seems to slight problem with it in the present state. 

For example, for the demo at http://troolee.github.io/gridstack.js/demo/nested.html when the outer grid is resized with mouse, it works great (and the inner grid items retains their relative positions and sizes). But when the *browser* is resized to be small, all widgets (including those inside at the inner levels) collapse to single vertical line (thus no more respecting the inner-outer hierarchy).

In the vertical mode, only the outer widget should become part of vertical line of widgets, while the "inner widgets still keep their relative positions and size" inside it. I believe the purpose of outer widget is to protect the inner levels from loosing their relative positions. 

If the present behavior (of making all equal width irrespective of inner-outer) is intention and by design, then at least there should be an *option* (something like, "retain inner structure" or "protect inner grid") which when set to true, the inner grids will retain their relative position and size irrespective of their parents.

troolee:
You can achieve that behaviour by disabling one-column mode for nested grids.

PR #1976, opened 2022-06-13
fix #418 Use of inline styles doesn't play nice with Content Security Policy

### LINKED ISSUE #418: Use of inline styles doesn't play nice with Content Security Policy
gridstack.js seems to use inline styles extensively. We would like to disable inline styles via CSP's style-src to avoid the security risk they present. Are there any plans for gridstack.js to avoid the use of inline-styles?

shaurya-sisodia:
fix #418
### Description
* updating styles directly on HTMLELement style description rather than updating dynamic stylesheets
* removing logic related to dynamic stylesheets as they aren't required and violate CSP strict style-src rule
* removed UTs related to Dynamic stylesheets

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -1198,22 +1198,7 @@ export class GridStack {
 
       // these are done once only
       Utils.updateStyleOnElements(prefix, {'min-height': cellHeight+cellHeightUnit});
-      // content margins
-      let top: string = this.opts.marginTop + this.opts.marginUnit;
-      let bottom: string = this.opts.marginBottom + this.opts.marginUnit;
-      let right: string = this.opts.marginRight + this.opts.marginUnit;
-      let left: string = this.opts.marginLeft + this.opts.marginUnit;
-      let content = `${prefix} > .grid-stack-item-content`;
-      let placeholder = `.${this.opts._styleSheetClass} > .grid-stack-placeholder > .placeholder-content`;
-      Utils.updateStyleOnElements(content, {top, right, bottom, left});
-      Utils.updateStyleOnElements(placeholder, {top, right, bottom, left});
-      // resize handles offset (to match margin)
-      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-ne`, {right});
-      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-e`, {right});
-      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-se`, {right, bottom});
-      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-nw`, {left});
-      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-w`, {left});
-      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-sw`, {left, bottom});
+      this._updateElementChildrenStyling(undefined, prefix);
     }
 
     // apply position styling on all dirty node's elements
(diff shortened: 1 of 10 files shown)


src/gridstack.ts
```diff
@@ -226,8 +221,8 @@ export class GridStack {
   protected _ignoreLayoutsNodeChange: boolean;
   /** @internal */
   public _gsEventHandler = {};
-  /** @internal */
-  protected _styles: GridCSSStyleSheet;
+  /** @internal max height of the gridstack*/
+  public _max: number;
```
adumesny, line 225: does it need to be public ? if so '_' isn't the right prefix, and rename to curRow or something (need to see how it's used)

src/gridstack.ts
```diff
@@ -240,6 +235,8 @@ export class GridStack {
   protected _extraDragRow = 0;
   /** @internal true if nested grid should get column count from our width */
   protected _autoColumn?: boolean;
+  /** @internal calculates height for a given row */
+  private _getHeight: (row: number) => string;
```
adumesny, line 239: make this a proper TS method, not var you assign to function later.

src/gridstack.ts
```diff
@@ -338,6 +335,8 @@ export class GridStack {
 
     this._setStaticClass();
 
+    this._getHeight = (rows: number) => (rows * (this.opts.cellHeight as number)) + this.opts.cellHeightUnit;
```
adumesny, line 338: see above

src/gridstack.ts
```diff
@@ -1209,47 +1198,38 @@ export class GridStack {
...
-    if (!this._styles) {
-      let id = 'gridstack-style-' + (Math.random() * 100000).toFixed();
-      // insert style to parent (instead of 'head' by default) to support WebComponent
-      let styleLocation = this.opts.styleInHead ? undefined : this.el.parentNode as HTMLElement;
-      this._styles = Utils.createStylesheet(id, styleLocation);
-      if (!this._styles) return this;
-      this._styles._id = id;
-      this._styles._max = 0;
+    if (!this._max) {
+      this._max = 0;
```
adumesny, line 1202: doesn't prevent if (!this._max) {...} clause, so likely incorrect placement.

src/gridstack.ts
```diff
@@ -1209,47 +1198,38 @@ export class GridStack {
...
+      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-e`, {right});
+      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-se`, {right, bottom});
+      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-nw`, {left});
+      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-w`, {left});
+      Utils.updateStyleOnElements(`${prefix} > .ui-resizable-sw`, {left, bottom});
     }
 
+    // apply layout styles on all widgets
+    const widgets: HTMLElement[] = Utils.getElements(prefix);
+    for (const widget of widgets) {
```
adumesny, line 1226: Utils.getElements(prefix).forEach(w => Utils.updatePositionStyleOnWidget(w, this._getHeight));

src/utils.ts
```diff
@@ -3,404 +3,401 @@
...
-        ret[key] = Utils.cloneDeep(obj[key]);
-      }
-    }
-    return ret;
-  }
-}
-
-// list of fields we will skip during cloneDeep (nested objects, other internal)
-const skipFields = ['_isNested', 'el', 'grid', 'subGrid', 'engine'];
+ import { GridStackElement, GridStackNode, GridStackOptions, numberOrString, GridStackPosition, GridStackWidget } from './types';
```
adumesny, line 6: indent all left so we can see actual changes...

adumesny (commented):
thank you for looking into this. Have  a few comments and I will need to pull it locally later to test as well.

shaurya-sisodia:
Hi @adumesny 

Thank you so much for such quick feedback.

I have incorporated the review comments. I have a couple of questions. I have fixed the wrong indent that I committed.
Please have a look. 

Regards,
Shaurya

src/gridstack.ts
```diff
@@ -1305,6 +1280,7 @@ export class GridStack {
     if (n.y !== undefined && n.y !== null) { el.setAttribute('gs-y', String(n.y)); }
     if (n.w) { el.setAttribute('gs-w', String(n.w)); }
     if (n.h) { el.setAttribute('gs-h', String(n.h)); }
+    Utils.updatePositionStyleOnWidget(el, this._getHeight);
```
adumesny, line 1283: I would pass n as that is the source of truth, not the gs-*** attributes

src/utils.ts
```diff
@@ -114,46 +114,42 @@ export class Utils {
...
-    } else if (typeof sheet.insertRule === 'function') {
-      sheet.insertRule(`${selector}{${rules}}`);
-    }
+  /** update position style on element */
+  static updatePositionStyleOnWidget(el: HTMLElement, getHeight:(rows: number) => string): void {
+    const h = Utils.toNumber(el.getAttribute('gs-h'));
+    const minH= Utils.toNumber(el.getAttribute('gs-min-h'));
+    const maxH = Utils.toNumber(el.getAttribute('gs-max-h'));
+    const y = Utils.toNumber(el.getAttribute('gs-y'));
+    el.style.height = getHeight(h);
```
adumesny, line 149: I would pass GridStackPosition instead.
and you are setting the height(s) and top only - what about left and widths ? seem one sided...

src/utils.ts
```diff
@@ -114,46 +114,42 @@ export class Utils {
...
-    }
-    if (!parent) {
-      // default to head
-      parent = document.getElementsByTagName('head')[0];
-      parent.appendChild(style);
-    } else {
-      parent.insertBefore(style, parent.firstChild);
+  /** update CSS style on elements with given selector */
+  static updateStyleOnElements(selector: string, styles: {[props: string]: string |  string[] }): void {
+    const elements = Utils.getElements(selector);
```
adumesny, line 119: Utils.getElements(selector).forEach(el => Utils.addElementStyle(el, styles));

src/utils.ts
adumesny: also I would not set properties (like min/max) that are not set

src/utils.ts
```diff
@@ -114,46 +114,42 @@ export class Utils {
...
     }
-    return style.sheet as CSSStyleSheet;
   }
 
-  /** removed the given stylesheet id */
-  static removeStylesheet(id: string): void {
-    let el = document.querySelector('STYLE[gs-style-id=' + id + ']');
-    if (el && el.parentNode) el.remove();
+  /** update CSS style on element */
+  static addElementStyle(el: HTMLElement, styles: {[props: string]: string |  string[] }): void {
```
adumesny, line 126: if (!(styles instanceof Object)) return; // not sure this even needed if (!styles) is enough
for (const key in styles) {
  if (!styles.hasOwnProperty(key)) return;
  const style = styles[key];
  if (Array.isArray(style)) { // why do we need this ?
    style.forEach(val => el.style[key] = val);
  } else {
    el.style[key] = style;
  }
}

adumesny (commented):
also your code will update all the elements styles when one of them changes, which will be very inefficient (prev code only did incremental stylesheet update as maxRow was reached).

without deleting gridstack.css (or at least any widget location/size info) and gridstack-extra.css you can't tell if you have handled all cases (I doubt it). Are you ONLY removing JS code that create inline styles, but still allowing static styles sheets (with dynamic class setting). I don't know enough about CSP to tell how far we would need to go...

this will need very different architecture.

shaurya-sisodia:
Hi @adumesny 

CSP strict style rule prohibits the following:

- Inserting rule using CSSStyleSheet.insertRule() or CSSStyleSheet.addRule()
- Appending the an element with inline style in the dom: Eg- El.appendChild(`<div style="width:100px"></div>`);
- Appending dynamically generated CSSStyleSheets into the DOM.

That's why I had to change the implementation such that we set styles on each element. 
In my architecture, we have to update Styles on each element because we have to take into account that the position on multiple elements might have changed too besides the CSS style calculation itself.

We can use the old architecture and still make it CSP compliant for all browsers except SAFARI. 
As written here in the following issue: 
I have come across [replaceSync accepts inline CSS which is not CSP compliant · Issue #98 · WICG/construct-stylesheets (github.com)](https://github.com/WICG/construct-stylesheets/issues/98)

We can use CSSStyleSheet.replaceSync() instead of insertRule() to update styles on dynamic style sheet. We can also instead of appending this dynamic stylesheet can use
`document.adoptedStyleSheets = document.adoptedStyleSheets.concat(sheet);`
 
 
 We can use old architecture with adoptedStyleSheets and replaceSync is all the browsers and new architecture in case of SAFARI. Or ask applications to use precalculated stylesheets in case of safari similar to [IE 8 Support](https://github.com/iongroup/gridstack.js/blob/v0.2.5-branch/README.md#ie8-support).

adumesny:
so we have a mix of css rules + classes for thing that don't change (eg column based layout) but attributes for height stuff...nthat would satisfy CSP but looks odd. Also we do NOT need to update all elements as we know when a node is dirty and needs to be updated (we already track that) so the call to generate styles need to taken out as it was done only of we added rows (or cellHeight/margin changed). Your logic isn't correct.

also keep in m id you changed break some of the demos so there are issues there - just playing with two.html shows issues dragging/resizing around. https://gridstackjs.com/demo/two.html is fine while your version using demo/two.html is not

shaurya-sisodia:
Hi @adumesny 

I am aware of the demos breaking and have fixed the changes locally and will commit shortly.

Also so we can use the architecture but update styles on elements only when they are dirty?

shaurya-sisodia:
Hi @adumesny 

I have fixed the code and incorporated the review you sent. The demo is also working perfectly.  We are updating styles on dirty ndes only

shaurya-sisodia:
@adumesny  Please review

adumesny:
this is likely not needed (and quite a risky change) given `nonce` support we now have.
https://github.com/gridstack/gridstack.js/pull/2229

I would like to close it - we can always get back to it from 'won't fix' issue listed.

PR #1989, opened 2022-06-28
Use Rollup to package gridstack.js as an ES Module

justinjxzhang:
### Description
This PR uses Rollup to package Gridstack.js as an ES module - in my use case I was unable to work with this as a CommonJS module.

### Checklist


justinjxzhang:
I've been digging into this more and found something that needs your input as it would be a breaking change.

The current builds don't use ES Modules but instead packaged as UMD. When moving to ES Modules it'll change how GridStack will need to be referenced - though this is primarily an issue within non-framework-based codebases.

For example in the "Getting Started" section of project website the JS boils down to

```html
<script src="node_modules/gridstack/dist/gridstack-h5.js"></script>
<script type="text/javascript">
  var items = [
    {content: 'my first widget'}, // will default to location (0,0) and 1x1
    {w: 2, content: 'another longer widget!'} // will be placed next at (1,0) and 2x1
  ];
  var grid = GridStack.init();
  grid.load(items);
</script>
```

with `GridStack` being in the global scope. ES Modules would require it to be shifted to this:

```html
<script type="module">
  import { GridStack } from 'node_modules/gridstack/dist/gridstack-h5.js';
  var items = [
    {content: 'my first widget'}, // will default to location (0,0) and 1x1
    {w: 2, content: 'another longer widget!'} // will be placed next at (1,0) and 2x1
  ];
  var grid = GridStack.init();
  grid.load(items);
</script>
```

Primary differences being the `script` now needs a `type` of `module`, and the import of GridStack needs to be inside that script and will no longer be available on the global scope.

I suspect this could cause issues for other users in this use case - would it then be better to simply have a separate `dist/esm` folder (similar to how the `es5` version is distributed) so that it doesn't break existing uses?

justinjxzhang:
@adumesny 

I've gone and made the ESM build separate from the existing ES6 CommonJS build, like how the ES5 build is separate. This is to ensure the smallest impact for users that are already using the library, but still provide the ability to include this in projects that require ES Modules.

As for the sizes, it appears a little bigger than the "standard" ES6 build, but similar to the size of the ES5 build. There are some gains and losses on both sides when you compare file-by-file, but overall they seem quite similar (looking at around 700KB for everything - h5, jQuery variant, jQuery + jQuery UI).

I did try to migrate the ES6 and ES5 builds over to Rollup so there was only one bundler but kept getting stuck around the packaging convention - for some reason it either didn't work or wanted to be called as `GridStack.GridStack.init`. Almost definitely misconfiguration on my side and something that could be explored in the future, but for now (while it's not great), I think two bundlers is kind of ok if that's alright with you.

Source maps should be there as well for debugging.

adumesny:
thanks for the update. I' swamped doing a big change to mouse event + touch for mobile so won't get to look for a couple weeks...

adumesny:
I don't see the benefit of using rollup, I can easily change to ES6/ES2020 for webpack target and will do that instead (my Angular app complais it's not ES module for optimization as well).

adumesny:
no need for rollup, I just need to change tsconfig.json module to ES6/ES2020 and done!
we still get a single .js for browsers (webpack umd or commonjs, no es) but that's ok I think.

adumesny:
well Flat ESM might be what we want after all... https://angular.io/guide/angular-package-format#flattening-of-es-modules

adumesny:
aaahhh, I was trying to modify this review (rebase on latest and tweaks) but instead ended up deleting index (wanted to revert but 1 change) and creating my own branch locally... now 'm getitng these runtime build errors (windows)

```js
./src/gridstack.ts 39 bytes [built] [code generated] [1 error]

ERROR in ./src/gridstack.ts
Module build failed (from ./node_modules/ts-loader/index.js):
Error: Debug Failure. Expected C:/depot/gridstack.js/tsconfig.json === C:\depot\gridstack.js\tsconfig.json. 
    at attachFileToDiagnostic (C:\depot\gridstack.js\node_modules\typescript\lib\typescript.js:17717:11)
    at attachFileToDiagnostics (C:\depot\gridstack.js\node_modules\typescript\lib\typescript.js:17746:36)
    at Object.parseJsonText2 [as parseJsonText] (C:\depot\gridstack.js\node_modules\typescript\lib\typescript.js:28669:41)
    at parseJsonText (C:\depot\gridstack.js\node_modules\typescript\lib\typescript.js:27781:19)
    at parseConfigFileTextToJson (C:\depot\gridstack.js\node_modules\typescript\lib\typescript.js:35667:28)
    at Object.readConfigFile (C:\depot\gridstack.js\node_modules\typescript\lib\typescript.js:35664:41)
    at getConfigFile (C:\depot\gridstack.js\node_modules\ts-loader\dist\config.js:18:31)
    at successfulTypeScriptInstance (C:\depot\gridstack.js\node_modules\ts-loader\dist\instances.js:74:58)
    at getTypeScriptInstance (C:\depot\gridstack.js\node_modules\ts-loader\dist\instances.js:42:12)
    at Object.loader (C:\depot\gridstack.js\node_modules\ts-loader\dist\index.js:15:67)
```

demo/esmodule.html
```diff
@@ -0,0 +1,88 @@
...
+        </div>
+      </div>
+    </div>
+    <div class="col-sm-12 col-md-10">
+      <div class="grid-stack"></div>
+    </div>
+  </div>
+
+  <script type="module">
+    import { GridStack } from '../dist/esm/gridstack.js';
```
adumesny, line 56: how can you load that ? in Chrome I get:

Access to script at 'file:///C:/depot/gridstack.js/dist/gridstack.js' from origin 'null' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, isolated-app, chrome-extension, chrome, https, chrome-untrusted.

justinjxzhang: I got that too; ended up serving the directory from the root of the repo (i.e. `npx http-server .` in `/gridstack.js`)

PR #2005, opened 2022-08-03
Update readme.md

sacalata:
changes the sass code snippet for extra cols, using the / instead of math.div in favor of backwards compatibility for older versions of sass
(math.div is only supported in sass 1.3.3 and above)

### Description

### Checklist


adumesny:
thank you

PR #2021, opened 2022-08-24
update Grid's options when setting float

bhowell2-fleetio:
Sync the Grid's options with the float value.

I also set to check if the value actually changed before calling `triggerChangeEvent`, but maybe this is not desired?

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -790,8 +790,10 @@ export class GridStack {
    * enable/disable floating widgets (default: `false`) See [example](http://gridstackjs.com/demo/float.html)
    */
   public float(val: boolean): GridStack {
-    this.engine.float = val;
-    this._triggerChangeEvent();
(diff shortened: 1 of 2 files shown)


src/gridstack.ts
```diff
@@ -529,6 +529,7 @@ export class GridStack {
         delete o.alwaysShowResizeHandle;
       }
       Utils.removeInternalAndSame(o, GridDefaults);
+      o.float = this.engine.float
```
adumesny, line 532: no I would do it on float(true) API to set them both instead. user might never call save.

bhowell2-fleetio: @adumesny changed. I also checked if the value actually changed so that `_triggerChangeEvents` wasn't called when nothing changed, but I can see this maybe being undesired too. Lemme know.

adumesny:
thank you.

PR #2028, opened 2022-09-03
Updated acceptWidget's documentation

shibisuriya:
### Description
Updated acceptWidget's documentation.

### Checklist


adumesny:
thanks for the doc update, but it's rather verbose and you didn't fix the incorrect API. I will trim it back - don't think you need to spell an example out.

Issue #2054, opened 2022-09-28
Can't enter text in textarea/input


sbkobaidze:
Can't enter text on textarea/input 

Using gridstack.js - 6.0.2v
Browser: Google Chrome
OS: Windows 10

https://jsfiddle.net/sabakobaidze/sqv0acgx/2/
## Expected behavior
Should be able to write text when clicking textarea/input
## Actual behavior
Not being able to type when clicking on textarea/input

adumesny:
thank for posting.

adumesny:
fixed in 6.0.3 - don't forget to donate if you find this lib useful!

PR #2063, opened 2022-10-09
Can't enter text in textarea

adumesny:
### Description
* fix for #2054
* make sure to prevent mousedown handling over known list of tags (like jqueryui did)

### Checklist

### DIFF
--- a/demo/serialization.html
+++ b/demo/serialization.html
@@ -30,6 +30,7 @@
     });
     
     grid.on('added removed change', function(e, items) {
+      if (!items) return;
       let str = '';
       items.forEach(function(item) { str += ' (x,y)=' + item.x + ',' + item.y; });
       console.log(e.type + ' ' + items.length + ' items:' + str );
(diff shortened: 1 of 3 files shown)


Gezdy:
For @adumesny 
Element with [contentEditable="true"] has the same issue. So, suggested code completion:

```
// make sure we are not clicking on known object that handles mouseDown (TODO: make this extensible ?) #2054
const skipMouseDown = ['input', 'textarea', 'button', 'select', 'option'];
const name = e.target.nodeName.toLowerCase();
if (skipMouseDown.find(skip => skip === name))
        return true;

//for contenteditable
if(e.target.closest('[contenteditable="true"]')){        	
        return true;
}
```
Further away in the same function _mouseDown, just before returning

```
e.preventDefault();
dd_manager_1.DDManager.mouseHandled = true;
return true;
```

e.preventDefault() prevents blur event which occurs just after mousedown event.  
So, skipped elements here will continue to have focus when dragging.
Suggested solution:

```
if (document.activeElement){
     document.activeElement.blur();
}
e.preventDefault();
dd_manager_1.DDManager.mouseHandled = true;
return true;
```

Thx a lot for all your great work !

adumesny:
do you want to make a pull request instead ? thanks.

PR #2075, opened 2022-10-12
fix: document not defined error during SSR fixes #2073

### LINKED ISSUE #2073: Nuxt SSR support
Gridstack version 7
Nuxt with SSR

Steps to reproduce
1. Create a nuxt app
2. Add a gridstack.init
3. See the error
image

## Expected behavior
No errors

## Actual behavior
Errors

stephenjason89:
### Description
Added a check if window exist
If it does not exist then most likely it is in SSR and should return isTouch = false

### Checklist
- [ yes ] All tests passing (`yarn test`)


stephenjason89:
@adumesny Thank you for the wonderful package.
I believe mobile browser have the window property as well on the client side.
During SSR it will not be available thus making isTouch = false on the server side
When it is sent to the client side, which is the mobile browser, window will be available thus making isTouch = true

Issue #2106, opened 2022-11-15
`ui-resizable-autohide` not removed on mouseOver after upgrade from v5-jq to v7


dfenerski:
After an upgrade from the jq version of v5 to v7 a bug started occurring where the css class `ui-resizable-autohide`  was not removed on hover (mouseOver). After inspecting the issue a bit more I've found that in the `_mouseOver` handler, there is a check, and there is already a `resized` item which stops the function execution.

However, there is no other item currently being resized. The bug occurs after rapidly adding new items/removing old ones. I do not know the codebase well enough to investigate further and currently I am unable to provide a codepen or such.

Adding a simple check in the `_mouseOver` handler to see whether the currently resized item is actually `this` solved my issue. This change is unobtrusive and IMO helps.

I've added a PR. If this cannot be merged due to lack of repro steps/bad fix I will have to resort to `patch-package`.

Versions:
CR latest
GridStack v7.1.1

Looking forward to clarifying this

Thanks for all the amazing work (v7 is amazing) and BR

Dimitar

PR #2107, opened 2022-11-15
fix: ui-resizable-autohide not removed on hover

dfenerski:
See #2106 

### DIFF
--- a/src/dd-resizable.ts
+++ b/src/dd-resizable.ts
@@ -130,7 +130,7 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   protected _mouseOver(e: Event): void {
     // console.log(`${count++} pre-enter ${(this.el as GridItemHTMLElement).gridstackNode._id}`)
     // already over a child, ignore. Ideally we just call e.stopPropagation() but see https://github.com/gridstack/gridstack.js/issues/2018
-    if (DDManager.overResizeElement || DDManager.dragElement) return;
+    if (DDManager.overResizeElement === this || DDManager.dragElement) return;
     DDManager.overResizeElement = this;
     // console.log(`${count++} enter ${(this.el as GridItemHTMLElement).gridstackNode._id}`)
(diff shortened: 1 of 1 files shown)


src/dd-resizable.ts
```diff
@@ -130,7 +130,7 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   protected _mouseOver(e: Event): void {
     // console.log(`${count++} pre-enter ${(this.el as GridItemHTMLElement).gridstackNode._id}`)
     // already over a child, ignore. Ideally we just call e.stopPropagation() but see https://github.com/gridstack/gridstack.js/issues/2018
-    if (DDManager.overResizeElement || DDManager.dragElement) return;
+    if (DDManager.overResizeElement === this || DDManager.dragElement) return;
```
adumesny, line 133: yeah that is not the right fix - if we were already resizing, we should get a mouseOver, and if something else, we shouldn't make us the resizing widget.
the real fix is to figure why DDManager.overResizeElement is not being reset when dragging items fast or something...

dfenerski: Investigating this further showed that the problem lies with multiple instances of GridStack having the same element refs led to the problem. As this shouldn't be happening the problem really lies in the application code. 

Closing this and sorry for the disturbance.

BR

adumesny: | multiple instances of GridStack having the same element refs 

that shouldn't happen naturally, so there might still be a bug somewhere in the lib ? would be good to track where the issue lies.

dfenerski: The application was responsible for managing the instances, so we solved the issue by tweaking the logic there. I doubt it is a legit use case to create multiple instances with the *same* items so this one I guess is not in the lib.

kanikab15: Thank you @dfenerski. I faced the same issue and doing a grid.destroy() at the right place solved it.

PR #2162, opened 2023-01-13
improves new react demo example

ruslauz:
### Description
- Adds change handler to set new layout  when items have been resized or moved
- Adds buttons for adding new item dynamically, for extending the example
- Mentions one possible bug in GridStack related to layout changing when an item has been removed from a grid
- Adds usage of ref instead of an id property

### DIFF
--- a/demo/react-hooks-controlled-multiple.html
+++ b/demo/react-hooks-controlled-multiple.html
@@ -143,7 +143,6 @@
         <div style={{ display: 'flex' }}>
           <div style={{ display: 'flex', width: '50%' }}>
             <ControlledStack
-              id='gs1'
               items={items1}
               addItem={(item) => {
                 setItems1(items => [...items, item])
@@ -156,7 +155,6 @@
           </div >
           <div style={{ display: 'flex', width: '50%' }}>
             <ControlledStack
-              id='gs2'
               items={items2}
               addItem={(item) => {
                 setItems2(items => [...items, item])
--- a/demo/react-hooks-controlled-multiple.html
+++ b/demo/react-hooks-controlled-multiple.html
@@ -110,7 +110,7 @@
       // Also templating forces you to spell out the 12+ attributes GS supports (only x,w,w,h done below) instead of passing a option structure that supports everything
       // is not robust as things get added, and pollutes the DOM attr for default/missing entries, vs optimized code in GS.
       // ********************
-      <div style={{width: '100%', marginRight: '10px'}}>
+      <div style={{ width: '100%', marginRight: '10px' }}>
         <div className="grid-stack" ref={gridContainerRef}>
           {items.map((item, i) => {
             return (
--- a/demo/react-hooks-controlled-multiple.html
+++ b/demo/react-hooks-controlled-multiple.html
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <title>Gridstack.js React integration example</title>
   <link rel="stylesheet" href="demo.css" />
-  <link rel="stylesheet" href="../dist/gridstack-extra.css" />
+  <link rel="stylesheet" href="../dist/gridstack-extra.css"/>
   <script src="../dist/gridstack-all.js"></script>
 
   <!-- Scripts to use react inside html - DEVELOPMENT FILES -->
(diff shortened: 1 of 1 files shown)


demo/react-hooks-controlled-multiple.html
```diff
@@ -49,64 +50,56 @@ <h2>Controlled stack</h2>
...
-            // can't pass n directly even though similar structs as it has n.el.gridstackNode which gives JSON error for circular write.
-            addItem({id:n.id, x:n.x, y:n.y, w:n.w, h:n.h});
-          });
-        })
-        .on('removed', (ev, gsItems) => {
-          if (grid._ignoreCB) return;
-          gsItems.forEach(n => removeItem(n.id));
-        })
-
+          .on('added', (ev, gsItems) => {
```
adumesny, line 61: wish you didn't change the indentation - makes it hard to know what you changed (and who did what in the history). Please revert indent.

demo/react-hooks-controlled-multiple.html
```diff
@@ -49,64 +50,56 @@ <h2>Controlled stack</h2>
...
+            if (grid._ignoreCB) return;
+            // remove the new element as React will re-create it again (dup) once we add to the list or we get 2 of them with same ids but different DOM el!
+            // TODO: this is really not ideal - we shouldn't mix React templating with GS making it own edits as those get out of sync! see comment below @111.
+            gsItems.forEach(n => {
+              grid.removeWidget(n.el, true, false); // true=remove DOM, false=don't call use back!
+              // can't pass n directly even though similar structs as it has n.el.gridstackNode which gives JSON error for circular write.
+              addItem && addItem({ id: n.id, x: n.x, y: n.y, w: n.w, h: n.h });
```
adumesny, line 68: so only read change is addItem && - why would it ever be null ?

demo/react-hooks-controlled-multiple.html
```diff
@@ -49,64 +50,56 @@ <h2>Controlled stack</h2>
...
+          .on('removed', (ev, gsItems) => {
+            /* Looks like a bug in GridStack */
+            const dirtyNodes = grid.engine.getDirtyNodes();
+            if (dirtyNodes !== undefined && dirtyNodes.length !== 0) {
+              const newItems = grid.save(false);
+              changeItems && changeItems(newItems);
+            }
+
+            if (grid._ignoreCB) return;
+            gsItems.forEach(n => removeItem && removeItem(n.id));
```
adumesny, line 80: same

demo/react-hooks-controlled-multiple.html
```diff
@@ -49,64 +50,56 @@ <h2>Controlled stack</h2>
...
+
+          })
+          .on('change', (ev, gsItems) => {
+            const newItems = grid.save(false);
```
adumesny, line 84: that is not the best way to do this... it's going to set a new list and REBUILD EACH widget all over again when only a few might have changed AND really only the x,y,w,h might have changed so the structure values need to change but not trigger a react rebuild because the items are already at the correct location. Only client side structure needs syncing.

demo/react-hooks-controlled-multiple.html
```diff
@@ -49,64 +50,56 @@ <h2>Controlled stack</h2>
...
+          })
       } else {
         //
         // update existing grid layout, which is optimized to updates only diffs (will add new/delete items for examples)
         //
         const grid = gridRef.current;
         const layout = [];
         items.forEach((a) => layout.push(
-          refs.current[a.id].current.gridstackNode || {...a, el: refs.current[a.id].current}
+          refs.current[a.id].current.gridstackNode || { ...a, el: refs.current[a.id].current }
```
adumesny, line 94: again don't reformat.

demo/react-hooks-controlled-multiple.html
```diff
@@ -142,35 +135,42 @@ <h2>Controlled stack</h2>
...
+          <div style={{ display: 'flex', width: '50%' }}>
+            <ControlledStack
+              items={items1}
+              addItem={(item) => {
+                setItems1(items => [...items, item])
+              }}
+              removeItem={(id) => {
+                setItems1(items => items.filter(i => i.id !== id))
+              }}
+              changeItems={(items) => setItems1(items)}
```
adumesny, line 153: see comment above

adumesny (commented):
hey thanks for the contribution - see my comments though.

demo/react-hooks-controlled-multiple.html
ruslauz: Ok. Done!

demo/react-hooks-controlled-multiple.html
ruslauz: It was done to avoid errors in the case if `addItem === undefined` i.e. wasn't provided, but ok i have reverted it back

demo/react-hooks-controlled-multiple.html
ruslauz: Ok. Reverted

demo/react-hooks-controlled-multiple.html
ruslauz: Ok. Done

demo/react-hooks-controlled-multiple.html
ruslauz: Hi, thank you for comments. I have fixed indentation

demo/react-hooks-controlled-multiple.html
ruslauz: Do you mean creating a list with new object will force GridStack to rebuild each widget or React? Because as far as I know it is not the problem for React since ids which are use as keys were not changed hence React will not remount elements. Please correct me if i am wrong.

demo/react-hooks-controlled-multiple.html
```diff
@@ -68,45 +69,35 @@ <h2>Controlled stack</h2>
...
+        })
       } else {
         //
         // update existing grid layout, which is optimized to updates only diffs (will add new/delete items for examples)
         //
         const grid = gridRef.current;
-        const layout = [];
-        items.forEach((a) => layout.push(
+        const layout = items.map((a) => 
           refs.current[a.id].current.gridstackNode || {...a, el: refs.current[a.id].current}
```
ruslauz, line 92: Regarding this line, i see that you are adding el property to each item, I also had to do so when i was writing a React wrapper component to avoid creating a new widget instead of updating. And i had to ts-ignore it, since i got typescript error, because the ` load()` function expects `Widget[]` as a parameter. It looks like a typing error in `load()`.

demo/react-hooks-controlled-multiple.html
adumesny: did you actually debug react and made sure no new elements were created ?
I know in Angualr wrapper I did it doesn't as the id have not changed (what it keys off).

ruslauz: Yes, I did.

demo/react-hooks-controlled-multiple.html
adumesny: yeah, the problem is the id YOU provide is not guaranteed to there nor unqiue, so the lib cannot depend on it being there... that's why I have _id internally to know what they are. so el is used (if present, else creates a new item).
I suppose I could check if id is set and assume callee does the right thing....

if you look at the Angular wrapper, I get the gridstackNode (which has _id and .el) to load, else it uses widget for new items positions...
https://github.com/gridstack/gridstack.js/blob/master/demo/angular/src/app/gridstack.component.ts#L129
with
https://github.com/gridstack/gridstack.js/blob/master/demo/angular/src/app/gridstack-item.component.ts#L39

adumesny: I shoudl document that better... but again have 2 list is the wrong approach, where as https://github.com/gridstack/gridstack.js/tree/master/demo/angular/src/app Angular wrapper is the right way.

demo/react-hooks-controlled-multiple.html
```diff
@@ -68,45 +69,35 @@ <h2>Controlled stack</h2>
           });
         })
         .on('removed', (ev, gsItems) => {
+          /* Looks like a bug in GridStack */
```
adumesny, line 72: what do you mean ? 'removed' should only be called on item being removed | added, and change on existing ones changing values. all 3 cases need to update your list - which is why I don't like have 2 list to sync and is not the right way...

demo/react-hooks-controlled-multiple.html
ruslauz: I have made my own react wrapper component over gridstack with a custom hook which can make any component draggable and GridStackDraggable component to make any react component gridstack draggable. Do you wish to have a look on it?

demo/react-hooks-controlled-multiple.html
ruslauz: There is one case when you move one widget from one grid into another and when the widget is removed form the original grid, a relayout  occurs but the change event is not triggered despite the fact that some widgets changed their position, so i have to manually check this and call change handler to have the relevant layout

demo/react-hooks-controlled-multiple.html
adumesny: that would be great - if you could post something similar to the Angular component I created, that would help others as I don't use React. Then peolple can interate and improved on it.

demo/react-hooks-controlled-multiple.html
adumesny: that sounds like a bug - please post one. I just tried [two.html](https://gridstackjs.com/demo/two.html) dragging from the right side (where float=false) and item that would cause an update and indeed grid1 doesn't trigger a change event like it should!

PR #2178, opened 2023-01-21
feat(scroll): add possibility to disable auto scrolling

VincentMolinie:
### Description
Right now the scroll is way too fast. To prevent that I've implemented my own system of scrolling, but I need the possibility for the scroll to be disabled to make it work correctly 😅 

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -2181,7 +2181,9 @@ export class GridStack {
       if (node._temporaryRemoved) return; // handled by dropover
       let distance = ui.position.top - node._prevYPix;
       node._prevYPix = ui.position.top;
-      Utils.updateScrollPosition(el, ui.position, distance);
+      if (!this.opts.disableAutoScroll) {
+        Utils.updateScrollPosition(el, ui.position, distance);
+      }
 
       // get new position taking into account the margin in the direction we are moving! (need to pass mid point by margin)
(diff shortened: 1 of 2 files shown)


adumesny:
thank for adding this option - if it's really too fast, maybe it would be better to tweak the current code instead ?

VincentMolinie:
Thanks a lot for merging this 🙏

VincentMolinie:
Hey @adumesny,
Could you please release a patch 👼  ?

adumesny:
I will once I get a couple fixes... lots of overhead to release

adumesny:
actually this is where it should have been implemented (what jqeury-ui did) instead of adding a top option per grid

```ts
/** Drag&Drop dragging options */
export interface DDDragOpt {
  /** class selector of items that can be dragged. default to '.grid-stack-item-content' */
  handle?: string;
  /** default to 'body' */
  appendTo?: string;
  /** if set (true | msec), dragging placement (collision) will only happen after a pause by the user. Note: this is Global */
  pause?: boolean | number;
  /** default to `true` */
  // scroll?: boolean;
...
```

Can you please revert your change and updae the doc as well ? thanks.
But I would **MUCH** prefer fixing the root cause than adding a 'bandaid' - I see #2188 (I don't beleive seeing that on windows so maybe a mac only issue ? I haven't looked...

adumesny:
@VincentMolinie I would appreciate getting better scrolling in the lib if you can move your custom version here instead...

Issue #2185, opened 2023-01-23
Gridstack overrides the original target element


fredericrous:
## Subject of the issue

Gridstack overrides the original target element

## Your environment
* version of gridstack.js - 7.2.2-dev
* chrome / macos 12.3

## Steps to reproduce

see function initEvent in src/utils

## Expected behavior

I want to get the original target element as well as the gridstack one

## Use case

when I drop a card on a html element external to gridstack,
I want the event dragstop to tell me the html element I dropped it to

adumesny:
can you explain the use case and where it's different and wy it would matter ?
also have you considered touch event in your case on mobile and how that might affect it ? thanks.

adumesny:
| a trash bin and want to drag and drop cards from gridstack to the trash in order to remove them

that exact example is supported natively in GS - see https://gridstackjs.com/demo/two.html
(see removable: '.trash' option)

are you saying the native mouseup (which becomes a dragstop) event would have the trash as the target event ?
if we wanted to know what it was dropped on, maybe we can add that field rather than pass the raw initial event as that wouldn't be the same across browsers or mobile vs desktop....

fredericrous:
I am gonna try the removable option, thank you for pointing this out! 

I think it is convenient to have a property that gives the original event. it could be prefixed with `unsafe_originalEvent` for example.
While I agree having an event object being the same for every browser is super important, I am septic that it is a good thing to totally hide the original event. I'm just expressing a PoV, no point to argue. I might indeed not need it in the near future

I'm fine with closing the issue and pr for the moment

PR #2186, opened 2023-01-23
Feat: add original event to drag and mouse events

fredericrous:
use case:

when I drop a card on a html element external to gridstack,
I want the event dragstop to tell me the html element I dropped it to 

problem:

Gridstack overrides the original target element, see #2185

usage example:

```js
    grid.on(
      'dragstop',
      (event: GridStackDragMouseEvent<DragEvent>, el: GridItemHTMLElement) => {
        const { originalEvent } = event;
        console.log('dragstop', originalEvent.target);
    });
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -476,24 +476,31 @@ export class Utils {
     }
   }
 
-  public static initEvent<T>(e: DragEvent | MouseEvent, info: { type: string; target?: EventTarget }): T {
-    const evt = { type: info.type };
-    const obj = {
-      button: 0,
-      which: 0,
-      buttons: 1,
-      bubbles: true,
-      cancelable: true,
-      target: info.target ? info.target : e.target
-    };
-    // don't check for `instanceof DragEvent` as Safari use MouseEvent #1540
-    if ((e as DragEvent).dataTransfer) {
-      evt['dataTransfer'] = (e as DragEvent).dataTransfer; // workaround 'readonly' field.
+    public static initEvent<T extends DragEvent | MouseEvent>(e: DragEvent | MouseEvent, info: { type: string; target?: EventTarget }): T {
+        return {
+            type: info.type,
+            button: 0,
+            buttons: 1,
+            bubbles: true,
+            cancelable: true,
+            target: info?.target || e.target,
+            ...((e as DragEvent).dataTransfer && {
+                dataTransfer: (e as DragEvent).dataTransfer
+            } as any),
+            // keys
+            altKey: e.altKey,
+            ctrlKey: e.ctrlKey,
+            metaKey: e.metaKey,
+            shiftKey: e.shiftKey,
+            // point info
+            pageX: e.pageX,
+            pageY: e.pageY,
+            clientX: e.clientX,
+            clientY: e.clientY,
+            screenX: e.screenX,
+            screenY: e.screenY,
+        } as T;
     }
-    ['altKey','ctrlKey','metaKey','shiftKey'].forEach(p => evt[p] = e[p]); // keys
-    ['pageX','pageY','clientX','clientY','screenX','screenY'].forEach(p => evt[p] = e[p]); // point info
-    return {...evt, ...obj} as unknown as T;
-  }
 
   /** copies the MouseEvent properties and sends it as another event to the given target */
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -476,24 +484,32 @@ export class Utils {
...
-    const evt = { type: info.type };
-    const obj = {
-      button: 0,
-      which: 0,
-      buttons: 1,
-      bubbles: true,
-      cancelable: true,
-      target: info.target ? info.target : e.target
-    };
-    // don't check for `instanceof DragEvent` as Safari use MouseEvent #1540
```
adumesny, line 489: you removed both comments and workaround readonly - please add them back. otherwuise why would you cast to any which si typically a no-no

src/utils.ts
```diff
@@ -476,24 +484,32 @@ export class Utils {
...
+            type: info.type,
+            button: 0,
+            buttons: 1,
+            bubbles: true,
+            cancelable: true,
+            target: info?.target || e.target,
+            ...((e as DragEvent).dataTransfer && {
+                dataTransfer: (e as DragEvent).dataTransfer
+            } as any),
+            // keys
```
adumesny, line 498: not clear how that is better than the compact methods that were there before...

src/utils.ts
```diff
@@ -476,24 +484,32 @@ export class Utils {
...
+            metaKey: e.metaKey,
+            shiftKey: e.shiftKey,
+            // point info
+            pageX: e.pageX,
+            pageY: e.pageY,
+            clientX: e.clientX,
+            clientY: e.clientY,
+            screenX: e.screenX,
+            screenY: e.screenY,
+            originalEvent: e,
```
adumesny, line 510: so this IS the only real change you made. not clear to me what you have is cleaner - definitely more verbose.

fredericrous: yeah that's why I did 2 commits to try to separate the real change to the reformat.
I closed the PR for now because I don't have the correct typings.

src/utils.ts
fredericrous: this part might not be clearer indeed.  I can put that condition outside

The actual part that is not clear is the forEach section
imo .forEach shouldn't be used to modify a variable that is external to it's scope as there is reduce for that.
also regarding minification and interpretation, I'm not sure the use of strings is very efficient.

src/utils.ts
```diff
@@ -3,7 +3,15 @@
  * Copyright (c) 2021 Alain Dumesny - see GridStack root license
  */
 
-import { GridStackElement, GridStackNode, GridStackOptions, numberOrString, GridStackPosition, GridStackWidget } from './types';
+import {
```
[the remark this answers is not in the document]
fredericrous: prettier did it for me because it exceeded 100 char but I can put that back

src/utils.ts
fredericrous: what's your thought on how to implement this `originalEvent: e` ?

src/utils.ts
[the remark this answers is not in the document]
adumesny: well everyone had a different formatting settings, that's why it's best not to auto reformat other people code...

src/utils.ts
adumesny: .forEach is just looping, not really a function per say... we use it in many places to deal with list of attributes - more compact code wise and doesn't suffer copy/paste mismatch (left and right side). please restore, unless ...() is used to get all those (and then some).

src/utils.ts
adumesny: as I asked in he bug posted, I need to understand why it is needed... never been an issue so trying to udnerstand the use case.

src/utils.ts
fredericrous: I found literature on why using .forEach for mutating data outside of the scope of the function you pass to it is unpure and should be avoided:

https://www.bennadel.com/blog/2992-mutating-an-array-during-foreach-iteration-in-javascript.htm

basically, using a regular for loop is safer. A functional programming approach is to use reduce.. which is more complicated, but you could create a helper function or use something like [pick](https://lodash.com/docs/4.17.15#pick) from lodash

PR #2192, opened 2023-02-01
fix(scroll): move the option to disable autoscroll

VincentMolinie:
### Description

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -2181,7 +2181,7 @@ export class GridStack {
       if (node._temporaryRemoved) return; // handled by dropover
       let distance = ui.position.top - node._prevYPix;
       node._prevYPix = ui.position.top;
-      if (this.opts.draggable.scroll === false) {
+      if (this.opts.draggable.scroll) {
         Utils.updateScrollPosition(el, ui.position, distance);
       }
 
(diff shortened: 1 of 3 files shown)


src/gridstack.ts
```diff
@@ -2181,7 +2181,7 @@ export class GridStack {
       if (node._temporaryRemoved) return; // handled by dropover
       let distance = ui.position.top - node._prevYPix;
       node._prevYPix = ui.position.top;
-      if (!this.opts.disableAutoScroll) {
+      if (this.opts.draggable.scroll === false) {
```
adumesny, line 2184: you mean
if (this.opts.draggable.scroll !== false) {...}

you also should update types.ts
gridDefaults: GridStackOptions = {...
draggable: { handle: '.grid-stack-item-content', appendTo: 'body', scroll: true },

and update the doc mentioning that new option...

VincentMolinie: Done 😄

src/gridstack.ts
```diff
@@ -2181,7 +2181,7 @@ export class GridStack {
       if (node._temporaryRemoved) return; // handled by dropover
       let distance = ui.position.top - node._prevYPix;
       node._prevYPix = ui.position.top;
-      if (this.opts.draggable.scroll === false) {
+      if (this.opts.draggable.scroll) {
```
adumesny, line 2184: !== false 
still better in case we fail to init that var (I know it's it in the defaul now). I'll update as I need to a release for work :)

adumesny:
thank you. any chance you might put in your scrolling speed fix in there instead ?

PR #2207, opened 2023-02-04
fixes incorrect order after changing items height

### LINKED ISSUE #2206: Incorrect order of items if a height of one item has been changed manually
## Subject of the issue
In case if you are changing a height of one item by changing the `h` property for instance by changing a state in react app you can face an incorrect order of items, i.e. the item standing next to the changed item is moved to the end of the list.

screencast 2023-02-04 21-57-02

## Your environment
* version of gridstack.js - 7.2.3
* OS - MacOs  13.0
* Browser - Chrome Version 108.0.5359.98 (Official Build) (arm64)

I managed to fix it locally by changing the third parameter of the `_fixCollisions` method on the 671 line of the `gridstack-engine.ts` file from `collide` to `undefined` forcing it to define collisions once again and to start moving items from the last to the first in the list

gridstack-engine ts — Frontend 2023-02-04 22-14-13

fixed

I can send a PR if it is a bug and not the feature

ruslauz:
### Description
The PR is fixing the next [issue](https://github.com/gridstack/gridstack.js/issues/2206)
fix #2206 

### Checklist


adumesny:
checked in the correct fix - behaves like update() instead.

PR #2218, opened 2023-02-22
Resolved: Grid engine.nodes count is incorrect after removing item.

XiaoHongYong:
### Description
Vue will reset items in order if no 'key' property is set. So, the element and id won't be matched between vue and GridStack.
So, the solution is to add key property.


adumesny:
thanks for fixing vue example!

PR #2233, opened 2023-03-17
docs: slack invite link on Readme

fredericrous:
Let users join slack from the link provided in the Readme

the link of the invite is already in the template when we create an issue, but users that have failed a first time to connect to slack won't always think that the invite link works

Most dev communities are on discord now, most ppl are not used to slack anymore. At least I'm not 🤷‍♂️


adumesny:
thank you!

PR #2239, opened 2023-03-25
Removed static css dependencies for grid-stack widgets. Dynamic styles…

tamil-arasu849:
… will be created based on the width and left custom style attributes.

So no need to create sttaic css files baed on the grid column.

This change will elimiate the need to create static css files for any gridcolumn. By this the grid can have any number of columns, we can increase and decrease dynamically as we wish


src/gridstack.scss
```diff
@@ -88,20 +88,8 @@ $animation_speed: .3s !default;
...
-    }
-  }
-
-  &.grid-stack-1>.grid-stack-item {
-    min-width: 100%;
-    &[gs-w='1'] { width: 100%; }
-    &[gs-x='1'] { left: 100%; }
-    &[gs-min-w='1'] { min-width: 100%; }
-    &[gs-max-w='1'] { max-width: 100%; }
+    width: calc(calc(100% / var(--grid-column)) * var(--gs-w));
```
adumesny, line 91: why do you have this ? we don't set width or left and nothing column based should be here with your cxhanges anyway,  line 8 $gridstack-columns: should be removed.

src/gridstack.ts
```diff
@@ -374,6 +374,7 @@ export class GridStack {
     this._updateStyles();
     if (this.opts.column != 12) {
```
adumesny, line 375: this probably should be done for 12. not sure why we ever special cased it.

src/gridstack.scss
adumesny: also if you are using dynamic style var(--grid-column) there is a performance tradeoff vs having it pre calculated. might be an issue with many grid items....

src/gridstack.scss
```diff
@@ -88,20 +88,8 @@ $animation_speed: .3s !default;
...
-      &[gs-min-w='#{$i}'] { min-width: math.div(100%, $gridstack-columns) * $i; }
-      &[gs-max-w='#{$i}'] { max-width: math.div(100%, $gridstack-columns) * $i; }
-    }
-  }
-
-  &.grid-stack-1>.grid-stack-item {
-    min-width: 100%;
-    &[gs-w='1'] { width: 100%; }
-    &[gs-x='1'] { left: 100%; }
-    &[gs-min-w='1'] { min-width: 100%; }
```
adumesny, line 103: where are you doing min/max then ?

src/gridstack.ts
```diff
@@ -1414,9 +1416,9 @@ export class GridStack {
 
   /** @internal call to write position x,y,w,h attributes back to element */
   protected _writePosAttr(el: HTMLElement, n: GridStackPosition): GridStack {
-    if (n.x !== undefined && n.x !== null) { el.setAttribute('gs-x', String(n.x)); }
+    if (n.x !== undefined && n.x !== null) { el.setAttribute('gs-x', String(n.x)); el.style.setProperty("--gs-x", String(n.x))}
```
adumesny, line 1419: I don't understand why setting an element style here is used with the rest of the code you have.... something missing.

adumesny (commented):
thanks, I like the idea of generating styles as needed, but dynamic var(--) styles have a performance issue I' not sure about for large grids...

src/gridstack.scss
tamil-arasu849: Yes need to handle min, max. I tried this idea and thought you would implement this in the lib.

src/gridstack.ts
tamil-arasu849: Will analyse it

src/gridstack.scss
tamil-arasu849: I think the performance overhead will minimal and would not be noticable, even for large number of items.

src/gridstack.ts
tamil-arasu849: I too 😅

src/gridstack.scss
tamil-arasu849: What do you mean by "nothing column based should not be there"? We should recompute grid items width and left position based on the updated column right? That's why I used the var(--grid-column), I set the attribute in the grid-container's div element and that will be recomputed here.

src/gridstack.ts
tamil-arasu849: > Will analyse it. I wish you could analyse, test this idea and implement this in the code base. The performance issues will not be noticeable  in most of the cases and we need not to worry about it as far as I know.

src/gridstack.scss
adumesny: ok I see now. but it's not ever just 'width' but gs-w 1/2/3/4/5/... you are assuming items are always 1x1 ?

src/gridstack.scss
adumesny: toughts wrong... I have a billions things to do, this isn't on top of my list. if you want this it needs to be a complete solution and test with >1k items before vs after.

src/gridstack.ts
adumesny: well I need proof & data, not 'think'

adumesny:
something like that is not a big deal

```scss
@use "sass:math";
$gridstack-columns: 150;
.grid-stack-{$gridstack-columns} > .grid-stack-item {

  min-width: 100%/$gridstack-columns;

  @for $i from 0 through $gridstack-columns {
    &[gs-w='#{$i}'] { width: (100% / $gridstack-columns) * $i; }
    &[gs-x='#{$i}'] { left: (100% / $gridstack-columns) * $i; }
  }
} 
```
actually I'm not sure why I have min/max-width today as that is handled in resize, not CSS anyway so 150x2

adumesny:
closing as not a complete solution not does it handle non size=1 items.

PR #2263, opened 2023-04-13
fix(drag): always put element under the mouse when dragging an element

VincentMolinie:
### Description
The position of elements dragged was badly calculated and resulted in weird behaviour in a lot of cases like:
- Disable automatic scroll, scroll page while dragging 
- Zoom out/in using the `scale` css property and try to drag a component

### Checklist

### Issues fixed
https://github.com/gridstack/gridstack.js/issues/1275

### DIFF
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -52,6 +52,7 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   protected static originStyleProp = ['transition', 'pointerEvents', 'position', 'left', 'top', 'minWidth', 'willChange'];
   /** @internal pause before we call the actual drag hit collision code */
   protected dragTimeout: number;
+  protected _originalMousePositionInsideElement: { x: number; y: number; };
 
   constructor(el: HTMLElement, option: DDDraggableOpt = {}) {
     super();
@@ -199,6 +200,10 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
       } else {
         delete DDManager.dropElement;
       }
+      if (node) {
+        const rect = this.el.getBoundingClientRect();
+        this._originalMousePositionInsideElement = { x: e.clientX - rect.left, y: e.clientY - rect.top };
+      }
       this.helper = this._createHelper(e);
       this._setupHelperContainmentStyle();
       const ev = Utils.initEvent<DragEvent>(e, { target: this.el, type: 'dragstart' });
@@ -280,6 +285,7 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
     // style.cursor = 'move'; //  TODO: can't set with pointerEvents=none ! (done in CSS as well)
     style.width = this.el.offsetWidth + 'px';
     style.height = this.el.offsetHeight + 'px';
+
     style.willChange = 'left, top';
     style.position = 'fixed'; // let us drag between grids by not clipping as parent .grid-stack is position: 'relative'
     this._dragFollow(e); // now position it
@@ -316,17 +322,14 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal updates the top/left position to follow the mouse */
   protected _dragFollow(e: DragEvent): void {
     const style = this.helper.style;
-    const containerElement = Utils.getPositionContainerElement(this.helper.parentElement);
-    const containerBoundingClientRect = containerElement.getBoundingClientRect();
     const { scaleX, scaleY } = Utils.getScaleForElement(this.helper);
 
+
     // Position the element behind the mouse
-    const node = (this.el as GridItemHTMLElement).gridstackNode || {};
-    const x = (e.clientX - containerBoundingClientRect.left - (node._originalMousePositionInsideElement?.x || 0)) / scaleX;
-    const y = (e.clientY - containerBoundingClientRect.top - (node._originalMousePositionInsideElement?.y || 0)) / scaleY;
+    const x = (e.clientX  - (this._originalMousePositionInsideElement?.x || 0)) / scaleX;
+    const y = (e.clientY  - (this._originalMousePositionInsideElement?.y || 0)) / scaleY;
     style.left = `${x}px`;
     style.top = `${y}px`;
-    console.log('isVisible', this.helper !== null)
   }
 
   /** @internal */
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -323,10 +323,11 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   protected _dragFollow(e: DragEvent): void {
     const style = this.helper.style;
     const { scaleX, scaleY } = Utils.getScaleForElement(this.helper);
-    // when an element is scaled, the helper is positioned relative to it's parent, so we need to remove the extra offset
-    const containementRect = this.helperContainment.getBoundingClientRect();
-    const offsetX = scaleX === 1 ? 0 : containementRect.left;
-    const offsetY = scaleY === 1 ? 0 : containementRect.top;
+    const transformParent = Utils.getContainerForPositionFixedElement(this.helper);
+    const transformParentRect = transformParent.getBoundingClientRect();
+    // when an element is scaled, the helper is positioned relative to the first transformed parent, so we need to remove the extra offset
+    const offsetX = transformParentRect.left;
+    const offsetY = transformParentRect.top;
 
     // Position the element under the mouse
     const x = (e.clientX - offsetX - (this._originalMousePositionInsideElement?.x || 0)) / scaleX;
@@ -350,14 +351,20 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal TODO: set to public as called by DDDroppable! */
   public ui(): DDUIData {
     const containmentEl = this.el.parentElement;
+    const scrollElement = Utils.getScrollElement(this.el.parentElement);
     const containmentRect = containmentEl.getBoundingClientRect();
     const offset = this.helper.getBoundingClientRect();
     const { scaleX, scaleY } = Utils.getScaleForElement(this.helper);
 
+    // When an element is inside a scrolled element, the boundingClientRect will return the position of the element minus the scroll.
+    const parentPositionIncludingScroll = containmentEl === scrollElement
+      ? { top: containmentRect.top + scrollElement.scrollTop, left: containmentRect.left + scrollElement.scrollLeft }
+      : { top: containmentRect.top, left: containmentRect.left };
+
     return {
       position: { // Current CSS position of the helper as { top, left } object
-        top: (offset.top - containmentRect.top) / scaleY,
-        left: (offset.left - containmentRect.left) / scaleX,
+        top: (offset.top - parentPositionIncludingScroll.top) / scaleY,
+        left: (offset.left - parentPositionIncludingScroll.left) / scaleX,
       }
       /* not used by GridStack for now...
       helper: [this.helper], //The object arr representing the helper that's being dragged.
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -202,7 +202,7 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
       }
       if (node) {
         const rect = this.el.getBoundingClientRect();
-        this._originalMousePositionInsideElement = { x: e.clientX - rect.left, y: e.clientY - rect.top };
+        this._originalMousePositionInsideElement = { x: s.clientX - rect.left, y: s.clientY - rect.top };
       }
       this.helper = this._createHelper(e);
       this._setupHelperContainmentStyle();
@@ -323,11 +323,14 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   protected _dragFollow(e: DragEvent): void {
     const style = this.helper.style;
     const { scaleX, scaleY } = Utils.getScaleForElement(this.helper);
+    // when an element is scaled, the helper is positioned relative to it's parent, so we need to remove the extra offset
+    const containementRect = this.helperContainment.getBoundingClientRect();
+    const offsetX = scaleX === 1 ? 0 : containementRect.left;
+    const offsetY = scaleY === 1 ? 0 : containementRect.top;
 
-
-    // Position the element behind the mouse
-    const x = (e.clientX  - (this._originalMousePositionInsideElement?.x || 0)) / scaleX;
-    const y = (e.clientY  - (this._originalMousePositionInsideElement?.y || 0)) / scaleY;
+    // Position the element under the mouse
+    const x = (e.clientX - offsetX - (this._originalMousePositionInsideElement?.x || 0)) / scaleX;
+    const y = (e.clientY - offsetY - (this._originalMousePositionInsideElement?.y || 0)) / scaleY;
     style.left = `${x}px`;
     style.top = `${y}px`;
   }
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -24,15 +24,6 @@ export interface DDDraggableOpt {
   drag?: (event: Event, ui: DDUIData) => void;
 }
 
-interface DragOffset {
-  left: number;
-  top: number;
-  width: number;
-  height: number;
-  offsetLeft: number;
-  offsetTop: number;
-}
-
 type DDDragEvent = 'drag' | 'dragstart' | 'dragstop';
 
 // make sure we are not clicking on known object that handles mouseDown
@@ -47,8 +38,6 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
 
   /** @internal */
   protected mouseDownEvent: MouseEvent;
-  /** @internal */
-  protected dragOffset: DragOffset;
   /** @internal */
   protected dragElementOriginStyle: Array<string>;
   /** @internal */
@@ -193,6 +182,11 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
         this._callDrag(e);
       }
     } else if (Math.abs(e.x - s.x) + Math.abs(e.y - s.y) > 3) {
+      let node = (this.el as GridItemHTMLElement)?.gridstackNode;
+      if (node) {
+        const rect = this.el.getBoundingClientRect();
+        node._originalMousePositionInsideElement = { x: e.clientX - rect.left, y: e.clientY - rect.top };
+      }
       /**
        * don't start unless we've moved at least 3 pixels
        */
@@ -207,7 +201,6 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
       }
       this.helper = this._createHelper(e);
       this._setupHelperContainmentStyle();
-      this.dragOffset = this._getDragOffset(e, this.el, this.helperContainment);
       const ev = Utils.initEvent<DragEvent>(e, { target: this.el, type: 'dragstart' });
 
       this._setupHelperStyle(e);
@@ -285,8 +278,8 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
     const style = this.helper.style;
     style.pointerEvents = 'none'; // needed for over items to get enter/leave
     // style.cursor = 'move'; //  TODO: can't set with pointerEvents=none ! (done in CSS as well)
-    style.width = this.dragOffset.width + 'px';
-    style.height = this.dragOffset.height + 'px';
+    style.width = this.el.offsetWidth + 'px';
+    style.height = this.el.offsetHeight + 'px';
     style.willChange = 'left, top';
     style.position = 'fixed'; // let us drag between grids by not clipping as parent .grid-stack is position: 'relative'
     this._dragFollow(e); // now position it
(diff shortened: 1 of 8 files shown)


adumesny:
thanks, will take a look. the title sounds scary, if all you wanted to fix was scrolling and scaling divs.

hosnyDev (approved):

src/gridstack.ts
```diff
@@ -2248,6 +2253,7 @@ export class GridStack {
       this._updateContainerHeight();
 
       let target = event.target as GridItemHTMLElement;// @ts-ignore
+      console.log('node', node.x, node.y, ui.position.left, ui.position.top)
```
adumesny, line 2256: need to comment/remove this

src/utils.ts
```diff
@@ -349,6 +349,18 @@ export class Utils {
     }
   }
 
+  static getContainerElement(el: HTMLElement): HTMLElement {
```
adumesny, line 352: might need to rename it better: getPositionContainer() or something. also can replace with while() loop rather than recursive...

src/utils.ts
```diff
@@ -529,6 +541,20 @@ export class Utils {
     (target || e.target).dispatchEvent(simulatedEvent);
   }
 
+  public static getNewXYPosition(event: MouseEvent, element: GridItemHTMLElement) {
+    const containerElement = Utils.getContainerElement(element.parentElement);
+    const containerBoundingClientRect = containerElement.getBoundingClientRect();
+    const boundingClientRect = element.getBoundingClientRect();
+    const scaleX = boundingClientRect.width / element.offsetWidth;
```
adumesny, line 548: woudl that work is scaling in set multiple times on different parents and not just immediate one ?

adumesny (commented):
so looks like you are handling parent div scale (what about general transform including translate ?) for dragging only, what about  for sizing do we need the same ?

src/utils.ts
VincentMolinie: Yes it will :) !

src/utils.ts
```diff
@@ -529,6 +541,20 @@ export class Utils {
...
+  public static getNewXYPosition(event: MouseEvent, element: GridItemHTMLElement) {
+    const containerElement = Utils.getContainerElement(element.parentElement);
+    const containerBoundingClientRect = containerElement.getBoundingClientRect();
+    const boundingClientRect = element.getBoundingClientRect();
+    const scaleX = boundingClientRect.width / element.offsetWidth;
+    const scaleY = boundingClientRect.height / element.offsetHeight;
+
+    // Position the element behind the mouse
+    const node = element.gridstackNode || {};
+    const x = (event.clientX - containerBoundingClientRect.left - (node._originalMousePositionInsideElement?.x || 0)) / scaleX;
```
adumesny, line 553: can't have a public Util APi call internal ._originalMousePositionInsideElement var from something, that breaks the internal boundary...

src/dd-draggable.ts
```diff
@@ -123,6 +123,12 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
 
   /** @internal call when mouse goes down before a dragstart happens */
   protected _mouseDown(e: MouseEvent): boolean {
+    let node = (this.el as GridItemHTMLElement)?.gridstackNode;
```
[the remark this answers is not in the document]
VincentMolinie: I tried but the thing is the element position has changed so it doesn't work inside the mouseMove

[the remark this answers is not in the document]
adumesny: the element shouldn't change until we drag >3 pixels, then we clone and make it float on top. right before that, we can generate the starting info you are doing here.
no point in doing that work if you are going to return or not drag enough...

src/dd-draggable.ts
```diff
@@ -184,6 +178,12 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
     // console.log(`${count++} move ${e.x},${e.y}`)
     let s = this.mouseDownEvent;
 
+    let node = (this.el as GridItemHTMLElement)?.gridstackNode;
```
adumesny: no, you wan tot move this to 203

this.dragging = true;
DDManager.dragElement = this;
... then your code

VincentMolinie:
I'm still working on it, I'll let you know as soon as I fixed all the issues

hosnyDev:
@VincentMolinie @adumesny  please can u fix this issue ASAP 😊

VincentMolinie:
Everything should be fixed now :)

VincentMolinie:
> @VincentMolinie @adumesny please can u fix this issue ASAP 😊

Yes don't worry that was my top priority 😄

VincentMolinie:
> @VincentMolinie @adumesny please can u fix this issue ASAP 😊

Yes don't worry that was my top priority 😄
It's just not that easy to fix ^^

hosnyDev:
@VincentMolinie appreciated TYT "D

VincentMolinie:
Well don't need to take my time, as I say I did fix every issues I could find 👼

VincentMolinie:
@adumesny it's back to you 😄

adumesny:
I pulled your changes locally, and demo/offset.html no longer works... can't trade scaling for offset.

VincentMolinie:
Good catch @adumesny , sorry about that. I fixed those issue and I added a demo for the scale so we can test it easily

src/dd-draggable.ts
```diff
@@ -189,6 +178,11 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
         this._callDrag(e);
       }
     } else if (Math.abs(e.x - s.x) + Math.abs(e.y - s.y) > 3) {
+      let node = (this.el as GridItemHTMLElement)?.gridstackNode;
+      if (node) {
+        const rect = this.el.getBoundingClientRect();
+        node._originalMousePositionInsideElement = { x: e.clientX - rect.left, y: e.clientY - rect.top };
```
adumesny, line 184: techinically this should use s.clientX (starting down event) so you are not 3 pixel + off from where you really clicked. if you move very fast that may be >3 pixels...

adumesny: also I don't like having it above the section comment as if this was special... move it down to where the offset was calculated.

src/dd-draggable.ts
```diff
@@ -319,15 +312,16 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
...
-    // }
     const style = this.helper.style;
-    const offset = this.dragOffset;
-    style.left = e.clientX + offset.offsetLeft - containmentRect.left + 'px';
-    style.top = e.clientY + offset.offsetTop - containmentRect.top + 'px';
+    const containerElement = Utils.getPositionContainerElement(this.helper.parentElement);
+    const containerBoundingClientRect = containerElement.getBoundingClientRect();
+    const { scaleX, scaleY } = Utils.getScaleForElement(this.helper);
+
+    // Position the element behind the mouse
```
adumesny, line 320: // Position the element under the mouse
is the proper english term.

src/dd-resizable.ts
```diff
@@ -277,10 +280,11 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
 
   /** @internal constrain the size to the set min/max values */
   protected _constrainSize(oWidth: number, oHeight: number): Size {
-    const maxWidth = this.option.maxWidth || Number.MAX_SAFE_INTEGER;
-    const minWidth = this.option.minWidth || oWidth;
-    const maxHeight = this.option.maxHeight || Number.MAX_SAFE_INTEGER;
-    const minHeight = this.option.minHeight || oHeight;
+    const { scaleX, scaleY } = Utils.getScaleForElement(this.el);
```
adumesny, line 283: const o = this.option;
then reduce code after to just
const maxWidth = o.maxWidth ? o.maxWidth * scaleX : Number.MAX_SAFE_INTEGER;

src/utils.ts
```diff
@@ -349,6 +349,18 @@ export class Utils {
...
 
+  static getPositionContainerElement(el: HTMLElement): HTMLElement {
+    if (!el) return null;
+
+    const style = getComputedStyle(el);
+
+    if (style.position === 'relative' || style.position === 'absolute' || style.position === 'fixed') {
+      return el;
+    } else {
+      return this.getPositionContainerElement(el.parentElement);
```
adumesny, line 360: there is no 'this' in static methods. use return Utils.getPositionContainerElement(...)
also no need to have else {} (a bit more compact, although less readable)

src/utils.ts
```diff
@@ -529,6 +541,20 @@ export class Utils {
     (target || e.target).dispatchEvent(simulatedEvent);
   }
 
+  public static getScaleForElement(element: HTMLElement) {
+    let elementToGetTheScaleFrom = element;
+
+    // Check if element is visible, otherwise the width/height will be of 0
+    while (!elementToGetTheScaleFrom.offsetParent) {
```
adumesny, line 548: should check for parent existing before else you could get js error.

src/utils.ts
```diff
@@ -529,6 +541,20 @@ export class Utils {
     (target || e.target).dispatchEvent(simulatedEvent);
   }
 
+  public static getScaleForElement(element: HTMLElement) {
+    let elementToGetTheScaleFrom = element;
```
adumesny, line 545: why re-assing ? you never use 'element' again. just shorten to el and use it below...

src/utils.ts
adumesny: check for el.parentElement being undefined

src/utils.ts
```diff
@@ -3,7 +3,7 @@
  * Copyright (c) 2021 Alain Dumesny - see GridStack root license
  */
 
-import { GridStackElement, GridStackNode, GridStackOptions, numberOrString, GridStackPosition, GridStackWidget } from './types';
+import { GridStackElement, GridStackNode, GridStackOptions, numberOrString, GridStackPosition, GridStackWidget, GridItemHTMLElement } from './types';
```
adumesny, line 6: you are not using (nor should the Util know about) GridItemHTMLElement - remove

adumesny:
your code is still completely breaking the main example demo/two.html (that and nested.html are the main 2 I always test against) at a minimum.

when you say you have fixed all issue, I'm not sure what you are using.
**Again, why do we need to suport div scaling ?** at this point it more risk breaking than is worth... Also you NEED to test mobile devices as I don't want have to test/fix mobile touch all over as well.

start drag make object jump to (0,0)
image

src/dd-draggable.ts
VincentMolinie:
```suggestion
    // Position the element under the mouse
```

adumesny:
I would strongly suggest you use the existing code but add scaling to the mix, rather than a complete re-write which has proven to be unreliable so far... twice, it broke very basic flows...

VincentMolinie:
I retested all the demo. now it should be good sorry about that 😥

DigitagDev:
@adumesny I work on a project where I drag items from a toolbar and drop them onto a grid. I even tried using the scale() function on the parent div to include a zoom effect on the area of my grid. Considering that several grids in the project need to have real scale, where a grid could occupy almost the entire screen or be a 200x200px grid with very small items, a zoom tool would be almost essential in my project. I'm mentioning this to clarify why I'm also trying to use the scale zoom. Unfortunately, browser support for this type of situation is only available for the entire page, not specifically for a div.

leemoria:
I also need scale，Have you some methods to resolve？@adumesny

adumesny:
I'll take a look this weekend...

adumesny:
ran out of time over the weekend (had bugs to fix) and I will be on PTO soon, so this will have to wait till I get back the weekend in ~2 weeks

VincentMolinie:
I found some issues, so you can delay your review 😞

VincentMolinie:
Okey I finally found the issue, and the reason why I had such a difference between the demo  and my own project.
The **Helper** is positioned in `fixed` when dragging. When having a position `fixed` it's positioned according to the viewport OR the first parent that has a `transform`. Which explain the weird position issue I had

So everything should be good now 🤞

src/dd-draggable.ts
```diff
@@ -193,6 +183,11 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
         this._callDrag(e);
       }
     } else if (Math.abs(e.x - s.x) + Math.abs(e.y - s.y) > 3) {
+      let node = (this.el as GridItemHTMLElement)?.gridstackNode;
+      if (node) {
+        const rect = this.el.getBoundingClientRect();
+        node._originalMousePositionInsideElement = { x: e.clientX - rect.left, y: e.clientY - rect.top };
```
adumesny, line 189: why do you have this code at all (shouldn't be on node, but the draggable class only) since you have same line 205. delete from GridStackNode

src/dd-draggable.ts
```diff
@@ -205,9 +200,12 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
       } else {
         delete DDManager.dropElement;
       }
+      if (node) {
+        const rect = this.el.getBoundingClientRect();
+        this._originalMousePositionInsideElement = { x: s.clientX - rect.left, y: s.clientY - rect.top };
```
adumesny, line 205: '_' is reserved for private vars, not protected. also rename to 'origRelativeMouse' and create based on this.mouseDownEvent otherwise you are off by 3+ pixels. ALSO don't really need that since we have this.mouseDownEvent (less vars to track)

src/dd-draggable.ts
```diff
@@ -205,9 +200,12 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
       } else {
         delete DDManager.dropElement;
       }
+      if (node) {
```
adumesny, line 203: no need for if (node)

src/dd-resizable.ts
```diff
@@ -311,23 +317,22 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
...
       left: this.originalRect.left,
-      top: this.originalRect.top - this.scrolled
+      top: (this.originalRect.top - this.scrolled)
     };
     const rect = this.temporalRect || newRect;
     return {
       position: {
-        left: rect.left - containmentRect.left,
-        top: rect.top - containmentRect.top
+        left: rect.left / scaleX,
```
adumesny, line 330: so this has a very different meaning - was relative position, and now it's not (and scaled). Thsi will affect user of those values...

src/types.ts
```diff
@@ -419,4 +419,6 @@ export interface GridStackNode extends GridStackWidget {
   _removeDOM?: boolean;
   /** @internal */
   _initDD?: boolean;
+  /** @internal original mouse position inside the element */
+  _originalMousePositionInsideElement?: GridStackPosition;
```
adumesny, line 423: remove

src/utils.ts
adumesny: you still have this.xyz in a static method!

src/utils.ts
```diff
@@ -360,6 +360,30 @@ export class Utils {
...
+
+  static getContainerForPositionFixedElement(el: HTMLElement | HTMLHtmlElement): HTMLElement {
+    if (el === document.documentElement) return el;
+
+    const style = getComputedStyle(el as HTMLElement);
+
+    if (style.transform && style.transform !== 'none') {
+      return el;
+    } else {
+      return this.getContainerForPositionFixedElement(el.parentElement);
```
adumesny, line 383: same

src/utils.ts
```diff
@@ -540,6 +564,24 @@ export class Utils {
     (target || e.target).dispatchEvent(simulatedEvent);
   }
 
+  public static getScaleForElement(element: HTMLElement) {
```
adumesny, line 567: (el: HTMLElement) then you don't need new local val.

adumesny:
still doens't behave the right way.

two.html the item jumps to cursor (0,0) when starting a new (helper) drag instead of being under the mouse
image
instead of https://gridstackjs.com/demo/two.html
image

also whe scaling (with zoom IN is much more optvious) you can see some weird wiggle animation where you must be changing the top & left position even though only scale bottom right happens. Didn't look at your code but it's not right.

https://github.com/gridstack/gridstack.js/assets/3496166/debc9eb1-3d50-45a5-8514-81ddfc84563c

src/utils.ts
VincentMolinie: Overriding a parameter is clearly not a way to go. It would make the code much less readable

src/dd-draggable.ts
VincentMolinie: I'm using `s` which is the `downEvent`.`el` can move so no, we need to keep that variable

VincentMolinie:
Everything is fixed @adumesny, thanks again for your review and sorry for the delay, I was in holidays ^^

src/dd-draggable.ts
```diff
@@ -345,51 +342,23 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
...
-      left: targetOffset.left,
-      top: targetOffset.top,
-      offsetLeft: - event.clientX + targetOffset.left - xformOffsetX,
-      offsetTop: - event.clientY + targetOffset.top - xformOffsetY,
-      width: targetOffset.width,
-      height: targetOffset.height
-    };
-  }
-
   /** @internal TODO: set to public as called by DDDroppable! */
```
adumesny, line 345: you've ppossibly changed the meaning of DDUIData here, yet this is used by DDDroppable according to comment, and I don't see changes in this review.... did you follow through and see how it is used and if this might be a breaking change ?

src/dd-resizable.ts
```diff
@@ -237,11 +237,14 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal */
   protected _getChange(event: MouseEvent, dir: string): Rect {
     const oEvent = this.startEvent;
+    const containerElement = Utils.getPositionContainerElement(this.el.parentElement);
+    const containerRect = containerElement.getBoundingClientRect();
+
     const newRect = { // Note: originalRect is a complex object, not a simple Rect, so copy out.
```
adumesny, line 243: to be honest I don't know _getChange() is used anymore...

src/dd-resizable.ts
```diff
@@ -277,28 +280,25 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
 
   /** @internal constrain the size to the set min/max values */
   protected _constrainSize(oWidth: number, oHeight: number): Size {
-    const maxWidth = this.option.maxWidth || Number.MAX_SAFE_INTEGER;
-    const minWidth = this.option.minWidth || oWidth;
-    const maxHeight = this.option.maxHeight || Number.MAX_SAFE_INTEGER;
-    const minHeight = this.option.minHeight || oHeight;
+    const { scaleX, scaleY } = Utils.getScaleForElement(this.el);
+    const o = this.option;
+    const maxWidth = o.maxWidth ? o.maxWidth * scaleX : Number.MAX_SAFE_INTEGER;
```
adumesny, line 285: if I recall this.options.maxWidth = n.maxW * cellWidth, so they still need to be scalled ?

src/dd-resizable.ts
```diff
@@ -277,28 +280,25 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
...
-      const { left, top } = containmentEl.getBoundingClientRect();
-      containmentRect = { left, top, width: 0, height: 0 };
-    }
     if (!this.temporalRect) return this;
-    Object.keys(this.temporalRect).forEach(key => {
-      const value = this.temporalRect[key];
-      this.el.style[key] = value - containmentRect[key] + 'px';
-    });
+    const { scaleX, scaleY } = Utils.getScaleForElement(this.el);
+    this.el.style.width = `${Math.floor(this.temporalRect.width / scaleX)}px`;
```
adumesny, line 298: why floor() and not round() ? if I recall you didn't have anyhting before and claimed all was working. what changed ?

src/utils.ts
```diff
@@ -360,6 +360,30 @@ export class Utils {
...
+    const style = getComputedStyle(el);
+
+    if (style.position === 'relative' || style.position === 'absolute' || style.position === 'fixed') {
+      return el;
+    } else {
+      return Utils.getPositionContainerElement(el.parentElement);
+    }
+  }
+
+  static getContainerForPositionFixedElement(el: HTMLElement | HTMLHtmlElement): HTMLElement {
```
adumesny, line 375: don't really need HTMLHtmlElement as you are not calling root methods...

Also class could be while() loop rather then recursive (might be easier to read).

src/utils.ts
```diff
@@ -360,6 +360,30 @@ export class Utils {
     }
   }
 
+  static getPositionContainerElement(el: HTMLElement): HTMLElement {
+    if (!el) return null;
+
+    const style = getComputedStyle(el);
```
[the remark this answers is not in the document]
VincentMolinie: It's from window not Utils 🤔

src/dd-draggable.ts
VincentMolinie: It's still the position of the helper 🤔

src/dd-resizable.ts
VincentMolinie: Yes it is. It is actually all the logic of the resize. It's called every time we resize

src/dd-resizable.ts
VincentMolinie: Yes because the scale apply to the grid so if the cell are 10px and you scale to 2 you will have actual cell of 20px

src/dd-resizable.ts
VincentMolinie: Nothing floor was working that's it. I can do round to

src/dd-resizable.ts
VincentMolinie: I didn't changed the meaning. This is the real position of the element. temporalRect has been updated to be at the right position

VincentMolinie:
Back to you @adumesny 😉

src/dd-resizable.ts
VincentMolinie: I replaced it by a round

src/utils.ts
```diff
@@ -540,6 +564,24 @@ export class Utils {
     (target || e.target).dispatchEvent(simulatedEvent);
   }
 
+  public static getScaleForElement(element: HTMLElement) {
+    let el = element;
+
+    // Check if element is visible, otherwise the width/height will be of 0
+    while (el && !el.offsetParent) {
```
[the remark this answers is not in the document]
adumesny: I know clientHeight\Width are 0 when invisible (makes sense), but didn't know about offsetParent (less obvious). I would sue that (which matches your comment)

adumesny:
still broken... while main demos seem fine (two.html and nested.html) offset.html doesn't work now.

https://github.com/gridstack/gridstack.js/assets/3496166/5ff57f14-22d2-475a-820f-37618adc7083

VincentMolinie:
The refacto did introduce a bug on the `getContainerForPositionFixedElement`

VincentMolinie:
@adumesny can you take a look at it 👼 ?

adumesny:
swamped at work and can't this weekend... I'll try to find time.

adumesny:
great, I tested a bunch of examples and mobile and it's working now. Will commit and go over the code again. thanks.

PR #2266, opened 2023-04-16
Esm rollup

adumesny:
this came from #1989


adumesny:
in the end the ouput of rollup is very much the same as built in tsc, so no bother... can't get roolup (needs package.json `"type": "module"`) or webpack to work at the same time right now

PR #2268, opened 2023-04-20
feat: added vue as submodule

shibisuriya:
### Description

### Checklist


adumesny:
I don't want a sperate module, it should be intergrated like I did for demo/Angular and published along GS.
Also I appreciate the effort but it feels weird to have dup documentation (you incorrectly state you must provide x,y for example (auto placement), use data when 'content' is what's built in, etc...

not clear how different from current vue example, and why you have introduced gs classes and tags (compare to what I did for angular - feels like more naturale extension to me).

PR #2274, opened 2023-04-23
Break: GridStackWidget.subGridOpts rename

adumesny:
### Description
* `GridStackOptions.subGrid` -> `GridStackOptions.subGridOpts`.
* We now have `GridStackWidget.subGridOpts` vs `GridStackNode.subGrid` (subclass) rather than try to merge the two at runtime since very different types...

always bugged me to have different runtime types, as it's error prone...

### Checklist


snapbakk:
Is this a breaking change when upgrading to the latest release ? I currently use version 7.2.3 and thankfully only the subGrid property needs to be renamed :+1:

PR #2378, opened 2023-06-29
add a class that defines a non removable grid item

DigitagDev:
### Description
An attribute 'decline' has been added to 'removableOptions' to deny the removal of a specific class. By default, it is set to 'grid-stack-non-removable'."

### Checklist


adumesny:
thanks for the feature!

PR #2385, opened 2023-07-08
Do not remove subgrid when updating widget content

adumesny:


src/gridstack.ts
```diff
@@ -1192,10 +1192,16 @@ export class GridStack {
...
-        if (sub && sub.innerHTML !== w.content) {
-          sub.innerHTML = w.content;
+        let toRemove = el.querySelectorAll('.grid-stack-item-content > :not(.grid-stack-nested, style)');
+        const subGrid = el.querySelector('.grid-stack-item-content > .grid-stack-nested')
+        const itemContent = el.querySelector('.grid-stack-item-content');
+        if (toRemove) {
+          toRemove.forEach(child => itemContent.removeChild(child));
         }
+        const tempEl = document.createElement("div");
+        tempEl.innerHTML = w.content;
```
adumesny, line 1202: I think you can do this instead: 

itemContent.innerHTML = w.content;
if (subGrid) itemContent.append(subGrid);

adumesny:
if (n.subGrid?.el) sub.appendChild(n.subGrid.el);
is all we need to add. will create CL

src/gridstack.ts
adumesny: nevermind, the CSS style gets removed, and re-adding it back doesn't seem to apply styles, so your solutionis better...

PR #2386, opened 2023-07-09
Do not remove subgrid when updating widget content

adumesny:


src/gridstack.ts
```diff
@@ -1192,10 +1192,16 @@ export class GridStack {
 
       // check for content changing
       if (w.content) {
-        let sub = el.querySelector('.grid-stack-item-content');
-        if (sub && sub.innerHTML !== w.content) {
-          sub.innerHTML = w.content;
+        let toRemove = el.querySelectorAll('.grid-stack-item-content > :not(.grid-stack-nested, style)');
```
adumesny, line 1195: Note: this doesn't work if you have content='foo' instead of content='<div>foo</div>' as in your jfiddle example

PR #2412, opened 2023-08-23
fitToContent fixes

### LINKED ISSUE #404: make widget auto-size height to fit content (no vertical scrollbar)
I want height as per inner container, i do not want any scroll bar and i want to put image inside  grid-stack-item-content and i want grid-stack-item height same as image Height,

i have tried but i do not get result, can please let me know it;s possible

adumesny:
### Description
more fix #404
* changing column size and manually resizing a widget will now call resizeToContent()
* added resizeToContentCB for app to override how to resize.
* Added `resizeToContentParent` to make generic code more robust.
* fixed some 1 column loading Nan issues

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ grid.printCount();
 
 ## Extend Engine
 
-You can now (5.1+) easily create your own layout engine to further customize you usage. Here is a typescript example
+You can now (5.1+) easily create your own layout engine to further customize your usage. Here is a typescript example
 
 ```ts
 import { GridStack, GridStackEngine, GridStackNod, GridStackMoveOpts } from 'gridstack';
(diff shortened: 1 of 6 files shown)


JonSohn:
Hi, I really needed this feature so I already tested it. However, I ran into a few issues that I wanted to share with you, as it might help:
1. right now the _fitToContent_ function only works if you have an additional container in the _grid-stack-item-content_ element. This is because of the _firstChild_ call in the _resizeToContent_ function which can be a Textnode or Comment due to spacing inside the dom. It is also addressed here in a sidenote. A possible solution would be to use _firstElementChild_ instead? This will still require a container, but it allows spacing between the elements.
2. I ran into an issue where a single element in the grid is not resized, even when configured to _fitToContent_. I traced it back to the _doContentResize_ function. The problem is caused by the iteration that the nodes go through in the engine. This is because within the loop each node is resized **and moved**, so during the iteration sometimes the nodes change their place in the array and are not processed when they are moved down in the array. A possible solution would be to call _resizeToContent_ with an array instead of individual nodes, or you could copy the array for iteration.

I hope I could help with this. If you have any questions, feel free to come back to me.

P.S: I don't usually work with GitHub. If this is the wrong place or I didn't follow something within the guidelines, please let me know.

adumesny:
great and thank you for finding the issues. I just released v9.0.0 but will incorporate those fixes in a review soon...

PR #2461, opened 2023-09-17
Remove added node from layout cache on drop event

### LINKED ISSUE #2394: Widget width reported incorrectly by save() after dragging from wider nested grid
## Subject of the issue
Widget width is incorrect in JSON produced by the `save()` method, if the given widget has been dragged from a nested grid that has more columns than the outer grid. The grid and widgets behave correctly, the problem is only with data generated by `save()`. Loading back this data into the grid will cause an incorrect state (since the JSON contains an impossible layout)

The `save()` method in `gridstack-engine.ts` takes the width value from `this._layouts` if present. The resized node's actual width seems correct, the incorrect value seems to come from `this._layouts`
https://github.com/gridstack/gridstack.js/blob/89ae9d6fd163137c5fde895ae14d1c5d2bb7a495/src/gridstack-engine.ts#L746C7-L746C7
image

## Your environment
* version of gridstack.js - 8.2.3
* which browser/OS - Firefox 116 / Chrome 114, Windows

## Steps to reproduce
1. Drag a widget from a grid with 2 columns into a nested grid with 3 columns
2. Resize the widget to be as wide as possible
3. Drag the widget back into the main grid from the nested grid
4. Resize the widget
The widget will have its width reported incorrectly in the JSON data generated by `save()`

https://jsfiddle.net/akosfekete/Lpv2au4w/11/
firefox_RNua2gRDXP

## Expected behavior
The width should be the same as it appears in the layout.

akosfekete:
### Description
Fix #2394 by removing dropped nodes from GridstackEngine's layout cache. 
I checked `1985_read_1_column_wrong_12.html` (the code relevant here has been added in that PR) and the behaviour is the same as before. Nodes are still positioned correctly after initialising in one column mode, and then enlarging the window (no matter how they are added)

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -2135,6 +2135,7 @@ export class GridStack {
         this._updateContainerHeight();
         this.engine.addedNodes.push(node);// @ts-ignore
         this._triggerAddEvent();// @ts-ignore
+        this.engine.removeNodeFromLayoutCache(node);
         this._triggerChangeEvent();
 
         this.engine.endUpdate();
--- a/src/gridstack-engine.ts
+++ b/src/gridstack-engine.ts
@@ -958,6 +958,17 @@ export class GridStackEngine {
     return this._layouts?.[column]?.findIndex(l => l._id === n._id) ?? -1;
   }
 
+  public removeNodeFromLayoutCache(n: GridStackNode) {
+    if (!this._layouts) {
+      return;
+    }
+    for (let i = 0; i < this._layouts.length; i++) {
+      let index = this.findCacheLayout(n, i);
+      if (index !== -1) {
+        this._layouts[i].splice(index, 1);
+      }
+    }
+  }
 
   /** called to remove all internal values but the _id */
   public cleanupNode(node: GridStackNode): GridStackEngine {


src/gridstack.ts
```diff
@@ -2135,6 +2135,7 @@ export class GridStack {
         this._updateContainerHeight();
         this.engine.addedNodes.push(node);// @ts-ignore
         this._triggerAddEvent();// @ts-ignore
+        this.engine.removeNodeFromLayoutCache(node);
```
adumesny, line 2138: this doesn't make sense to me as new grid target should not have that node to start with in it's cache.
instead line 2092 oGrid.engine.removeNodeFromLayoutCache(node); should be called to remove when LEAVING that grid.
I'll pull your changes and make the correct fix.

adumesny: nevermind I stand corrected. the issue is dropping an item that is larger into a grid - should it remember the width it wants to be and reloading should handle out of bound sizes (that is a bug) ? visually you are mathcing the wanted size but it w=2 outer grid vs w=3 inner grid. sicne it's a drag we visually match them (closest possible) so I'm ok with dropping the original width.

adumesny: the problem with your fix is that if an item wants to be w=3 (from sidebar with attributes values) but dropped into a 1 column, then you will forever loose the fact it wants to get widget if grid ever goes to 2,3,4 columns

the real fix is to make sure loading w=3 into a subgrip does the right thing - which IT DOES in v9.2.1 I tested with my example (based on your bug). So I will remove it.

PR #2481, opened 2023-09-30
v9.3.0-dev

adumesny:


Mitesh147:
Hello,

We noticed a strange behavior when loading gridstack dashboard in a touch-enabled device e.g., windows surface. when the component is loaded in such devices, it is showing "Resize" icons for every tiles by default and doesn't hide the icons on mouse events. Even drag and drop are not properly working.

We can also reproduce a similar behavior in a normal device. To reproduce, we were doing following steps -

1. Load gridstack dashboard demo page ( [https://gridstackjs.com/#demo] ) in firefox browser.
2. Click F12 to open Developer Tools.
3. Click on Responsive Design Mode icon in the top-right corner.
4. Refresh the page and you will see the exact issue.

It will show something like following screenshot
image

Any help would be appreciable.

Thanks,
Mitesh J.

adumesny:
resize handles show up by default on touch devices (to let you know which sides/corner support resizing), but didn't think about devices that support both.  change `alwaysShowResizeHandle: 'mobile'` to false to hide them.
Also this isn't a bug report. Slack channel for Q

don't forget to donate if oyu find this lib useful.

PR #2489, opened 2023-10-02
(fix): moved the `dropped` event handler call to a safer place

arnoudb:
(fix): In rare cases the call to trigger the `dropped` handler could lead to errors.

### Description
The error occurred in the `window.setTimeout` code right after. An example is when one executes some logic on the (droppedCB) event in Angular. If you reset the grid the item was dragged from (by resetting the items in an *ngFor) the grid crashed.

By executing the event in the `setTimeout` this problem is solved. It also makes logically sense as the event noting the drop is done is now fired after `this._prepareDragDropByNode(node);` is executed.

### Checklist

unfortunately i can't get the tests running (tried in the master branch )

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -2134,7 +2134,6 @@ export class GridStack {
           this.engine.cleanupNode(node); // removes all internal _xyz values
           node.grid = this;
         }
-        delete node.grid._isTemp;
         dd.off(el, 'drag');
         // if we made a copy ('helper' which is temp) of the original node then insert a copy, else we move the original node (#1102)
         // as the helper will be nuked by jquery-ui otherwise. TODO: update old code path
@@ -2162,17 +2161,26 @@ export class GridStack {
           subGrid.parentGridItem = node;
           if (!subGrid.opts.styleInHead) subGrid._updateStyles(true); // re-create sub-grid styles now that we've moved
         }
-        this._prepareDragDropByNode(node);
         this._updateContainerHeight();
         this.engine.addedNodes.push(node);// @ts-ignore
         this._triggerAddEvent();// @ts-ignore
         this._triggerChangeEvent();
-
         this.engine.endUpdate();
-        if (this._gsEventHandler['dropped']) {
-          this._gsEventHandler['dropped']({...event, type: 'dropped'}, origNode && origNode.grid ? origNode : undefined, node);
-        }
-        
+
+        // wait till we return out of the drag callback to set the new drag&resize handler or they may get messed up
+        window.setTimeout(() => {
+          // IFF we are still there (some application will use as placeholder and insert their real widget instead and better call makeWidget())
+          if (node.el && node.el.parentElement) {
+            this._prepareDragDropByNode(node);
+          } else {
+            this.engine.removeNode(node);
+          }
+          delete node.grid._isTemp;
+          if (this._gsEventHandler['dropped']) {
+            this._gsEventHandler['dropped']({...event, type: 'dropped'}, origNode && origNode.grid ? origNode : undefined, node);
+          }
+        });
+
         return false; // prevent parent from receiving msg (which may be grid as well)
       });
     return this;


src/gridstack.ts
```diff
@@ -2146,11 +2146,7 @@ export class GridStack {
         this.engine.addedNodes.push(node);// @ts-ignore
         this._triggerAddEvent();// @ts-ignore
         this._triggerChangeEvent();
-
         this.engine.endUpdate();
-        if (this._gsEventHandler['dropped']) {
```
adumesny, line 2151: callback here should work as line 2157 has comment about checking that we are here (and not deleted by user) so I need to better understand what crash you were seeing - calling after that statement doesn't make sense of that statement so something else needs to change as 2158 will always be true.

conceptually I don't see why CB can't be called after like you have, other than maybe doing all that work preparing a node that won't exists in the scene soon after...

arnoudb: Hi, thnx for the fast reply,

The situation can be described as the next in pseudo code:

```
protected items: [item1, item2, item3]

handleDropped() {
	this.items = [];
}

<gridstack id=“grid1”>
	<gridstack-item *ngFor="let item of items;
		<myItem item=“item” />
	</gridstack-item>
<gridstack />

<gridstack id=“grid2” (droppedCB)="handleDropped()"><gridstack />
```

Drag an item from grid 1 to grid 2:

```
Runtime error:
core.mjs:10183 ERROR TypeError: Cannot read properties of undefined (reading 'grid')
    at dd-gridstack.js:53:51
    at Array.forEach (<anonymous>)
    at DDGridStack.draggable (dd-gridstack.js:42:33)
    at GridStack._prepareDragDropByNode (gridstack.js:1982:16)
    at gridstack.js:1868:26
    at timer (zone.js:2367:41)
    at _ZoneDelegate.invokeTask (zone.js:402:31)
    at core.mjs:25945:55
    at AsyncStackTaggingZoneSpec.onInvokeTask (core.mjs:25945:36)
    at _ZoneDelegate.invokeTask (zone.js:401:60)
```

`_prepareDragDropByNode` is called after the `(droppedCB)` already fired the event signalling that it was complete. We cleared grid 1 immediately on the event. This interferes with the code in the `setTimeout`

If we emit the ‘dropped’ event in the `setTimeout` like in my PR we don’t get the runtime error and all works fine. Would save us some unneeded setTimeouts in our code and would make the lib more robust. Logic wise it makes (to me) more sense to emit a done event like ‘dropped’ when something is really done and not before the process of cleaning up.

I don't have enough knowledge on  the lib to see why the setTimeout is needed in the first place.

thnx!

arnoudb: Hi @adumesny,

I wondered if i can help out further to resolve this issue?  Do you need more information? Or maybe you can propose another way to fix this? I would like to help, but not sure what to do. We are using public api's so the crash should not happen. We could fork but that is definitely something we would like to avoid. We rather would like to help out in the official repo.

thnx!

adumesny: yes let's defintely fix it in the official repo - I was on vacation this week so catching up to a ton of work.

adumesny: | We cleared grid 1 immediately on the event
technically you clear the array right away, but ng needs next cycle to acutally update the dom ngFor loop (unless you also do that synchronously).

let me test why we have a timeout at all (comment doesn't have bug number and can't recall).

could you create a simple reproduceable case from the speudo code above so I can make sure to test that as well ?

arnoudb:
Hi, thnx for picking this up.

Unfortunately the new fix is not working for our project.

To see the new solution fail you can check out the next branch i made. Its a very simple demo.
https://github.com/arnoudb/gridstack.js/tree/drop-handler-bug-demo

see demo 8.

Tried fixing the source without the setTimeout like you did. Whatever i do, it doesn't work in our app. 
The initial solution i proposed (moving the call to the dropped handler inside the setTimeout) does work consistently.

Really hope you will reconsider applying the solution in the initial PR i made.

Thnx!

arnoudb:
example to show what happens
drag-drop-reset-items

arnoudb:
Hi, 

I rebased my original fix to resolve the conflicts. 
Hope you'll consider merging it.

cheers

adumesny:
since this is an angular (which I use) can you post a simple running example so I can understand the remaining issue and fix it in the lib (if that's the right place). I reather avoijd setTimeout() as hacks.

arnoudb:
Hi, i'm not sure what other example you would like to see? 

The example i gave above was written to do what you asked for. It shows that the grid still breaks with the alternative solution you implemented in 9.4.0. The example is for a specific case we use. You can checkout the branch and run it.

https://github.com/arnoudb/gridstack.js/tree/drop-handler-bug-demo

The same example works without error with the solution I proposed in this PR.

adumesny:
missed it. will have to take a look...

adumesny:
I don't know what you are trying to do removing all left side items when one of them get dragged, but you are doing it bunch of things the wrong way. I fixed your code and not getting any errors, but the item dropped is removed. If you really want this funky code you can do a setTimeout() in your cb instead

1. you're putting *ngFor on the grid instead of item!
```html
<gridstack [options]="demo8GridOpt">
        <gridstack-item *ngFor="let n of demo8GridElements1; trackBy: identify">
          <div>
            {{n.id}}
          </div>
        </gridstack-item>
      </gridstack>
```
2. don't need to set of options
3. ngFor loop is NOT recommended in my wrapper when dealing with 2 grids like this and ng tracking array and GS moving objects can be out of sync - hence the dropped item disapearing.

I went through the effort to show 2 grid+sidebar example which works. this is your issue.

arnoudb:
Hi, 

You are absolutely right about point one. That was a dumb mistake on my behalf.
I don't understand point 2, but it should be irrelevant to the demo I just wanted to give the grids minimal options.

You say:

> 'If you really want this funky code you can do a setTimeout() in your cb instead' 

That is what we already have and wanted to get rid of. The whole purpose of the starting the PR. I don't think the code is funky though. We have a complex data driven app with many grids that render complex widgets. These widgets have many inputs and outputs. We want the grids to just represent the models and use declarative code. Just be able to push new items whenever we want. So the model is then leading and the grid just a renderer. Hence our preference for the *ngFor. That worked great but only not in the case when dropping from another grid and then updating the models immediately. I'm still convinced there is a timing issue somewhere that happens because of our more complex code compared to the example.

The sad thing is that when i apply my proposed fix on top of 9.4.0 we can remove the `setTimeouts` in our code and everything works fine. When using 9.5.0 it completely breaks even without any changes in our code (thus keeping the `setTimeouts`). And i don't see an easy solution besides the one i gave.

So in the end, the result of my effort is not being able to update to future versions after 9.4.0 and nothing else changed for the better.

adumesny:
I spent more time on this already than I typically do already... if you want more support, expect to give a donation.

arnoudb:
Hi, i just got a message from my employer that you received the donation (Crunchr). I'm happy about that, as we really appreciate the project and all the effort you put in.

Hopefully we can explore this issue more and come to a solution that is really robust for *ngFor and multiple grids. 

We understand the limitations of that approach (the dropped items is re-created because we let ngFor in control), but that is something we accepted in favour of our code structure).

adumesny:
yes, thank you. as you say the issue with *ngFor is the dropped item is recreated by angular later - which is why **and tracking all GS edits** - is best to avoid and the angular wrapper I use and publish here does not use ngFor.
You will need to track add + remove AND any size editing made by the user to update your structure used by ngFor to be correct.

All do-able, but I really think you should use `gsCreateNgComponents` example and have each widget use angular templating nomrally, just just GS wrapper.

arnoudb:
Hi, I fully understand you want to get rid of the `setTimeout` in the lib. I don't like em either.
However fact is that is was there already before my PR and no-one seemed to be bothered by it.  Moving the `this._prepareDragDropByNode` in there (it may not be the lightest function) seems not a big deal as it's called  for 1 node on a drop action. So performance is not an issue i'd say. 

It would really help us if we could have at least one (minor release) with my fix. 
Is there a chance you are willing to do that? 

In the meantime I can work on / think about an alternative solution.

arnoudb:
After quite some experimentation i found out that there are 2 distinct issues:
The problem was located here:

```
        // wait till we return out of the drag callback to set the new drag&resize handler or they may get messed up
        window.setTimeout(() => {
          // IFF we are still there (some application will use as placeholder and insert their real widget instead and better call makeWidget())
          if (node.el && node.el.parentElement) {
            this._prepareDragDropByNode(node);
          } else {

            // --> you removed the if and this else. Now you always call:  this._prepareDragDropByNode(node);
            this.engine.removeNode(node);
          }
```

So my code relied on 2 things which are both true in my PR:

1.  `  this.engine.removeNode(node);` in the setTimeout to be executed which it always did on cross grid dropped. Probably because i updated the items for the *ngFor right away. which causes the `else` to be executed.
2.  that the 'dropped' event is dispatched afterwards. (the node is gone so no conflict for `*ngFor` to recreate)

In your fix 1 doesn't happen anymore. So i have to do it manually now.

I'm experimenting with the next which is quite ugly. Maybe there is a better way? For 9.4.0 i have:
```
@ViewChild('gridStack') gridStack?: GridstackComponent;

 ngAfterViewInit(): void {
    // monkeypatch grid
    this.gridStack?.grid?.on('dropped', (event, previousNode, newNode) => {
      console.log('remove the node, let ngFor create it');
      this.gridStack?.grid?.engine.removeNode(newNode);
      setTimeout(() => {
        console.log('dropped on grid is delayed ');
        this.gridStack?.droppedCB.emit({ event, previousNode, newNode });
      }, 10);
    });
``` 

For 9.5.1 i can get rid of the setTimeout which is nice

adumesny:
again use 9.5.1 as is. the code you are showing is gone from the lib and not coming back... I'm trying to help you with what any remaining issues there might be, not solving prev issues.

arnoudb:
You made it already pretty clear that you are not going to make any changes :-) 
I just tried to describe that it's now really clear what happens.  And I also wanted to let you know that i have got it all working (but with one patch). 

The question was basically: is there an cleaner way to patch the dropped event? 
But in the meantime i already found a better way. 

Anyway, this whole thing became a long exercise, but in the end helped me to remove lots of annoying setTimeouts in my code while keeping our code structure as we like. So the goal is met! Cheers.

PR #2495, opened 2023-10-07
Possible fix for React

Meierschlumpf:
### Description
In this pull request I've added a new option to the GridStackOptions. This option is called `disableRemoveNodeOnDrop` and it allows to define if a widget which is moved from one to another grid should be removed from the dom or just hidden with `display: none`.

This changes make not a lot of sense for normal users. But for users with a virtual dom it makes a ton of sense to not remove the item from the dom, 
especially when using state. By only hiding the item, the state will be able to remove the item from the dom.

I tried to run the tests, but they are not working for me.

### Checklist


adumesny:
sorry but this is a band aid option... as I mention in Angular wrapper doc, you should let GS move/reparent and add widgets around and not use ngFor kind of loop and instead of the addRemoveCB to create Angular components as needed, else let GS do it's thing.
This feels like an extension to that CB **if at all**

adumesny:
closing due to no reply and comment above.

Issue #2498, opened 2023-10-11
Animation issue after upgraded to 9.3.0


nhannv:
## Subject of the issue
I got the drag animation issue (I'm using framer-motion) after upgraded to 9.3.0, the issue also can be reproduced in gridstack's animation example which is using AniJS

## Your environment
* version of gridstack.js - 9.3.0
* which browser/OS - MacOS

## Steps to reproduce
The issue is occurring on Gridstack's animation example: https://gridstackjs.com/demo/anijs.html

## Expected behavior
It should smoothly when drag item.

adumesny:
@nhannv does this work in 9.2 ? if so another one for @VincentMolinie to take a look.

At this point (and drag offset not correct fix) I'm tempted to revert back mouse dragging as we're having so many problem with your new code Vincent, and the review itself took a lot of try. I don't want to test everything that can happen. **instead** you should have added scale to the existing code instead of trowing it all away...

nhannv:
@adumesny Yes, I can confirm the version 9.2 works fine

PR #2499, opened 2023-10-11
fix(scroll): prevent wrong offset when the scroll is on the html element

VincentMolinie:
### Description
Having scroll on the `html` element was increasing the offset because that element behave a bit differently, meaning that the scroll was included in the position of the html element.
So we had basically position like
```js
{ top: -150, left: 0 }
```
But if it's the `html` element in our case we were always expecting `0`

### Checklist

### Issues related
Fix https://github.com/gridstack/gridstack.js/issues/2491

### DIFF
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -328,9 +328,20 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
     //   containmentRect = { left, top };
     // }
     const style = this.helper.style;
-    const offset = this.dragOffset;
-    style.left = e.clientX + offset.offsetLeft - containmentRect.left + 'px';
-    style.top = e.clientY + offset.offsetTop - containmentRect.top + 'px';
+    const { scaleX, scaleY } = Utils.getScaleForElement(this.helper);
+    const transformParent = Utils.getContainerForPositionFixedElement(this.helper);
+    // We need to be careful here as the html element actually also includes scroll
+    // so in this case we always need to ignore it
+    const transformParentRect = transformParent === document.documentElement ? { top: 0, left: 0 } : transformParent.getBoundingClientRect();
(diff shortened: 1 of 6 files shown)


src/dd-draggable.ts
```diff
@@ -317,7 +317,9 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
     const style = this.helper.style;
     const { scaleX, scaleY } = Utils.getScaleForElement(this.helper);
     const transformParent = Utils.getContainerForPositionFixedElement(this.helper);
-    const transformParentRect = transformParent.getBoundingClientRect();
+    // We need to be careful here as the html element actually also includes scroll
+    // so in this case we always need to ignore it
+    const transformParentRect = transformParent === document.documentElement ? { top: 0, left: 0 } : transformParent.getBoundingClientRect();
```
adumesny, line 322: isn't this assuming here that the entire doc is scrollable, but it could be any parent div in between. I thought we had code to detect scrolled parent and take that into content...

VincentMolinie: yes actually you are right the issue is bigger than that. I will fix it 😄

VincentMolinie:
Okey I tested pretty much everything and I couldn't find issue. It should be good 🤞

adumesny:
thanks for this fix but #2498 remains and I have to revert all your changes for now as 9.3 is worst, until you can prove your entire new code path is better than what it replaced. **would have been much better to add scale to existing code instead of hunting/debugging all the issues you have now created**.

VincentMolinie:
This one is fixed too: https://github.com/gridstack/gridstack.js/issues/2498 but like I told you for the other one. I don't see how it is related. It seems more related to a PR that is 3 weeks old, not mine

adumesny:
going with https://github.com/gridstack/gridstack.js/pull/2515 instead of your re-write

PR #2500, opened 2023-10-11
lint: make linter more restrictive and add prettier

VincentMolinie:
### Description

### Checklist


src/dd-base-impl.ts
```diff
@@ -3,10 +3,12 @@
  * Copyright (c) 2021-2022 Alain Dumesny - see GridStack root license
  */
 
-export type EventCallback = (event: Event) => boolean|void;
+export type EventCallback = (event: Event) => boolean | void;
 export abstract class DDBaseImplement {
   /** returns the enable state, but you have to call enable()/disable() to change (as other things need to happen) */
-  public get disabled(): boolean   { return this._disabled; }
+  public get disabled(): boolean {
```
adumesny, line 9: I disagree with forcing things to be 3 lines for simple getter/setter.

src/dd-draggable.ts
```diff
@@ -112,7 +121,7 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   }
 
   public updateOption(opts: DDDraggableOpt): DDDraggable {
-    Object.keys(opts).forEach(key => this.option[key] = opts[key]);
+    Object.keys(opts).forEach((key) => (this.option[key] = opts[key]));
```
adumesny, line 124: another I disagree.

src/dd-draggable.ts
```diff
@@ -261,10 +270,13 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
       helper = Utils.cloneNode(this.el);
     }
     if (!document.body.contains(helper)) {
-      Utils.appendTo(helper, this.option.appendTo === 'parent' ? this.el.parentElement : this.option.appendTo);
+      Utils.appendTo(
+        helper,
```
adumesny, line 274: same, there shouldn't be a max width.

adumesny (changes requested):
I disagree with prettty default. All for const, space, comma, but don't go reformatting my code into n lines if I like it compact, or insert uneeded ()

if we can't prevent that, then pretty if off for me...

adumesny:
don't like pretty default.

Issue #1275, opened 2020-05-14
mouse drag out of sync with div scale


hzjux001:
When I change the target div's SCALE, just like "transform: scale(0.457292) translate(0px, 0px);",the movement or dragging in the gridstack is out of sync with the real mouse movement

adumesny:
you need to create a reproduceable test case if you want us to take a look. look at the jsfiddle template

hzjux001:
Because In many cases we need to zoom the page，
in the example https://gridstackjs.com/demo/advance.html,
I change the style in grid-stack div 
just like 
<div class="grid-stack grid-stack-instance-9799 grid-stack-animate ui-droppable" data-gs-animate="yes" data-gs-current-row="6" style="height: 460px;transform: scale(0.7, 0.7) translate(0px, 0px);">
and when we drag or resize by mouse,
the movement or dragging in the gridstack is out of sync with the real mouse movement

hzjux001:
The same is true even if you put the zoom style on the outer DIV

hzjux001:
Who can help me solve this problem,Please.

adumesny:
browser zooming works just fine, so not sure why you would manually scale the div in code...

if you are such in need of a fix, and don't want to fix it yourself, maybe you should consider donating $ to this effort.

hzjux001:
How can I donate

adumesny:
github page has donate links https://github.com/gridstack/gridstack.js

vishaldodiya:
@adumesny Is there any workaround for this? Like we can calculate position according to scale and pass it to gridstack.

adumesny:
unlikely to get fixed as not a priority and I don't understand the use case why you would scale a div (so many things break when scaling in my experience).

HTML5 drag&drop is a lot worse than the JQ version which seems only off by scale factor relative to parent (0,0)

adumesny:
fixed in next release.

adumesny:
will be in next 9.5 relase.

PR #2515, opened 2023-10-21
fix(scale): update drag & resize when the gridstack parent is scaled

elmehdiamlou:
### Description
When wanting to drag a grid item within a parent that is scaled using CSS transforms it doesn't work properly.

This PR serves as a solution for issue #1275 by determining the scaling factor applied to the parent to adjust the grid position and size. e.g. if the parent is scaled with scale(0.5), the scaling factor will be 2.

Additionally, here is a video demonstrating how this fix works in action.

https://github.com/gridstack/gridstack.js/assets/90270530/9e9677fc-008e-4b62-aa32-c123fc76b80d

### Checklist

### DIFF
--- a/demo/transform.html
+++ b/demo/transform.html
@@ -13,21 +13,25 @@
 <body>
   <div class="container-fluid">
     <h1>Transform Parent demo</h1>
-    <p>example where the grid parent has a translate(50px, 100px) <s>scale(0.5, 0.5)</s> (scale not working yet see #1275)</p>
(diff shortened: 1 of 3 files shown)


adumesny:
thank you for using existing code and just adding scale to it - this makes sure all existing path will continue working. wish I had not wasted so much time on @VincentMolinie review and frequent breakage....

VincentMolinie:
Glad I could help... I was just trying to help

PR #2530, opened 2023-11-10
fix(typings): make engine interface strict

Ruslan207:
### Description
add more strict interface for `GridStackEngine` class. It allows to extend engine (or make custom one) in projects with strict tsconfig

### Checklist


adumesny:
thanks

PR #2558, opened 2023-12-04
fix: remove style node when attached to parent node in a shadow root

confususs:
### Description

This is a fix for web components that use the shadow dom.

Basically, whenever a stylesheet is removed, `document.querySelector()` is used, and this doesn't work when gridstack is used within a shadow root. 

To fix this, we can make use of the `styleLocation`, which describes whether we're attaching the styles in a parent node or the head of a document

### Checklist
- [ ] All tests passing (`yarn test`) < I can't get tests working on `master` either :see_no_evil:

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -1489,7 +1489,8 @@ export class GridStack {
   protected _removeStylesheet(): GridStack {
 
     if (this._styles) {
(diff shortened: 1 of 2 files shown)


adumesny:
thank you.

PR #2573, opened 2023-12-12
Fix quickly drag elements between nested blocks

v1talii-dev:
### Description
If you quickly drag elements between nested blocks, duplicates of the same element appear.

### DIFF
--- a/src/dd-droppable.ts
+++ b/src/dd-droppable.ts
@@ -94,7 +94,7 @@ export class DDDroppable extends DDBaseImplement implements HTMLElementExtendOpt
 
     // make sure when we enter this, that the last one gets a leave FIRST to correctly cleanup as we don't always do
     if (DDManager.dropElement && DDManager.dropElement !== this) {
-      DDManager.dropElement._mouseLeave(e as DragEvent);
+      DDManager.dropElement._mouseLeave(e as DragEvent, true);
     }
     DDManager.dropElement = this;
 
@@ -108,7 +108,7 @@ export class DDDroppable extends DDBaseImplement implements HTMLElementExtendOpt
   }
 
   /** @internal called when the item is leaving our area, stop tracking if we had moving item */
-  protected _mouseLeave(e: MouseEvent): void {
+  protected _mouseLeave(e: MouseEvent, external = false): void {
     // console.log(`${count++} Leave ${this.el.id || (this.el as GridHTMLElement).gridstack.opts.id}`); // TEST
     if (!DDManager.dragElement || DDManager.dropElement !== this) return;
     e.preventDefault();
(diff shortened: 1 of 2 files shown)


src/gridstack.ts
```diff
@@ -2530,7 +2526,6 @@ export class GridStack {
     } else if (node._isExternal) {
       // item came from outside (like a toolbar) so nuke any node info
       delete node.el;
-      delete el.gridstackNode;
```
adumesny, line 2533: doesn't make sense to delete one and not the other... will have to debug the code for that particular case and make sure it doesn't break other cases..

v1talii-dev: I agree. I noticed errors in the console and problems with blocks not being deleted after dragging. Restored this function.

src/gridstack.ts
```diff
@@ -2111,18 +2111,14 @@ export class GridStack {
...
-        }
 
         // calculate the grid size based on element outer size
         helper = helper || el;
         let w = node.w || Math.round(helper.offsetWidth / cellWidth) || 1;
         let h = node.h || Math.round(helper.offsetHeight / cellHeight) || 1;
 
         // if the item came from another grid, make a copy and save the original info in case we go back there
-        if (node.grid && node.grid !== this) {
+        if (node.grid !== this) {
```
adumesny, line 2121: still doesn't feel right to me without debuging the issue - if node doesn't have grid, we go through that logic which doesn't feel right...

v1talii-dev: I agree with you. Now it seems to me that the problem is not in this place at all, but in calling the mouseEnter event in mouseOver. If you comment on this line as a definition of the source of the problem, then everything works correctly - https://github.com/gridstack/gridstack.js/blob/master/src/dd-droppable.ts#L135

v1talii-dev:
Redid PR. I discovered a bug of redundant event calling. When the "mouseenter" event was called, the "mouseleave" event was called inside that event. That is, they cleared the previous location of the block and showed a new one. The "mouseleave" event contains the logic for inserting a location into the parent (when we drag the block up into the parent). So, we got a situation where we dragged a field to a child, and the “mouseenter” event was triggered, in turn, “mouseleave” was called, and then “mouseenter” was called. Removed the redundant "mouseenter" step specifically for the program call (the behavior when moving the mouse was preserved).

src/dd-droppable.ts
```diff
@@ -120,7 +120,7 @@ export class DDDroppable extends DDBaseImplement implements HTMLElementExtendOpt
     }
     this.triggerEvent('dropout', ev);
 
-    if (DDManager.dropElement === this) {
+    if (!external && DDManager.dropElement === this) {
```
adumesny, line 123: DDManager.dropElement === this can be removed since on line 113

src/dd-droppable.ts
```diff
@@ -120,7 +120,7 @@ export class DDDroppable extends DDBaseImplement implements HTMLElementExtendOpt
...
       delete DDManager.dropElement;
```
adumesny, line 124: this should still happen. move to line 114.
rename the var fromParent and instead the next comment should be:
// if we're still over a parent droppable, send it an enter as we don't get one from leaving nested children if it's not already calling us
if (!fromParent) {
  let parentDrop: DDDroppable;
  let parent: DDElementHost = this.el.parentElement;
  ....

PR #2574, opened 2023-12-13
Allow cell height in cm and mm units

joostme:
### Description
Allow `cm` and `mm` units for cell height. I am using this to create a printable layout that's consistent across devices when printing on paper. The only way to achieve that is by using `cm` and `mm` units. This small PR allows the two additional CSS units.

### Checklist

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -219,7 +219,7 @@ export class Utils {
     if (typeof val === 'string') {
       if (val === 'auto' || val === '') h = 0;
       else {
-        let match = val.match(/^(-[0-9]+\.[0-9]+|[0-9]*\.[0-9]+|-[0-9]+|[0-9]+)(px|em|rem|vh|vw|%)?$/);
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -219,7 +219,7 @@ export class Utils {
     if (typeof val === 'string') {
       if (val === 'auto' || val === '') h = 0;
       else {
-        let match = val.match(/^(-[0-9]+\.[0-9]+|[0-9]*\.[0-9]+|-[0-9]+|[0-9]+)(px|em|rem|vh|vw|%)?$/);
+        let match = val.match(/^(-[0-9]+\.[0-9]+|[0-9]*\.[0-9]+|-[0-9]+|[0-9]+)(px|em|rem|vh|vw|%|cm|mm)?$/);
```
adumesny, line 222: do browser support cm ? and how would I convert this to px to do match (which I do for rem/em)

joostme: Browsers do support cm/mm. You can see the spec here: https://www.w3.org/TR/css-values-3/#absolute-lengths
There's also a conversion rate given for cm to px. I updated the `getCellHeight` method to accommodate for cm/mm units

adumesny:
thanks!

PR #2578, opened 2023-12-15
apply position scaling when acceptwidgets is enabled

Amdoun:
### Description
This fixes the issue i mentioned in https://github.com/gridstack/gridstack.js/issues/2575
Some of the code was added from https://github.com/gridstack/gridstack.js/commit/3c9bc4b1dc394201900c147f185c29ca5c892603 to the event handler for dragging when acceptWidgets is set to true (which is located in gridstack.ts and not in dd-draggable.ts)

### Checklist
- [x] Created tests which fail without the change (if possible) - No
- [ ] All tests passing (`yarn test`) - I could not run the tests, can someone try it on my PR?
- [x] Extended the README / documentation, if necessary - Not necessary

### DIFF
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -64,8 +64,6 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal pause before we call the actual drag hit collision code */
   protected dragTimeout: number;
   /** @internal */
-  protected transformReference: HTMLElement;
-  /** @internal */
   protected dragTransform: DragTransform = {
     xScale: 1,
     yScale: 1,
@@ -81,7 +79,6 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
     // get the element that is actually supposed to be dragged by
     let handleName = option.handle.substring(1);
     this.dragEl = el.classList.contains(handleName) ? el : el.querySelector(option.handle) || el;
-    this.transformReference = Utils.createTransformReferenceElement();
     // create var event binding so we can easily remove and still look like TS methods (unlike anonymous functions)
     this._mouseDown = this._mouseDown.bind(this);
     this._mouseMove = this._mouseMove.bind(this);
@@ -218,7 +215,6 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
       this.helper = this._createHelper(e);
       this._setupHelperContainmentStyle();
       this.dragTransform = Utils.getValuesFromTransformedElement(
-        this.transformReference,
         this.helperContainment
       );
       this.dragOffset = this._getDragOffset(e, this.el, this.helperContainment);
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -4,7 +4,7 @@
  */
 
 import { DDManager } from './dd-manager';
-import { Utils } from './utils';
+import { DragTransform, Utils } from './utils';
 import { DDBaseImplement, HTMLElementExtendOpt } from './dd-base-impl';
 import { GridItemHTMLElement, DDUIData } from './types';
 import { DDElementHost } from './dd-element';
@@ -33,11 +33,6 @@ interface DragOffset {
   offsetTop: number;
 }
 
-interface DragScaleReciprocal {
-  x: number;
-  y: number;
-}
-
 type DDDragEvent = 'drag' | 'dragstart' | 'dragstop';
 
 // make sure we are not clicking on known object that handles mouseDown
@@ -55,8 +50,6 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal */
   protected dragOffset: DragOffset;
   /** @internal */
-  protected dragScale: DragScaleReciprocal = { x: 1, y: 1 };
-  /** @internal */
   protected dragElementOriginStyle: Array<string>;
   /** @internal */
   protected dragEl: HTMLElement;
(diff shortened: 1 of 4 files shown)


src/gridstack.ts
```diff
@@ -2030,16 +2030,35 @@ export class GridStack {
...
+
     let onDrag = (event: DragEvent, el: GridItemHTMLElement, helper: GridItemHTMLElement) => {
       let node = el.gridstackNode;
       if (!node) return;
 
       helper = helper || el;
+      helper.appendChild(testEl);
+      const testElPosition = testEl.getBoundingClientRect();
+      helper.removeChild(testEl);
+      const dragScale = {
```
adumesny, line 2053: you will need the offset as well, which makes now 3 copies of that code so it should be extracted into a util method instead.

Amdoun: I moved some redundant logic for that to utils. I also now made it possible for elements dragged from outside to be scaled down when entering the grid, and scaled back up when they leave. Also their position were following the unscaled grid so I fixed that too. 
I am not sure which offset you mean but I did adjust the offset for the outside item's helpers when they enter the grid.

https://github.com/gridstack/gridstack.js/assets/23745606/43abe1e9-71a8-469e-915e-cfcb4b9bc26a

src/dd-draggable.ts
```diff
@@ -79,6 +81,7 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
     // get the element that is actually supposed to be dragged by
     let handleName = option.handle.substring(1);
     this.dragEl = el.classList.contains(handleName) ? el : el.querySelector(option.handle) || el;
+    this.transformReference = Utils.createTransformReferenceElement();
```
adumesny, line 84: is there a reason to save this as all ? we only need to calculate the scale/offset so no need to keep that element around (which then also needs to be freed

Have Utils.getValuesFromTransformedElement() create and delete it instead (internally)

src/utils.ts
```diff
@@ -554,6 +561,40 @@ export class Utils {
...
+      height: '1px',
+      zIndex: '-999999',
+    });
+    return transformReference;
+  }
+
+  /**
+   * can be used after setting the reference element from createTransformReferenceElement
+  */
+  public static getValuesFromTransformedElement(transformReference: HTMLElement, helper: HTMLElement): DragTransform {
```
adumesny, line 586: should called 'parent' (where the xform child is added)

adumesny (commented):
looks great, but more cleanup please.

src/dd-draggable.ts
Amdoun: it used to create a single instance of that div that could be reused for the next calls. i don't know what were the implications of creating this div and clearing it up on every single mousemove call. probably not much to worry about, so now it creates that element, gets the values, then deletes it.

on a side note, it turns out just calling parent.removeChild(transformReference) was not enough to delete the element: i had to call transformReference.remove() too

src/utils.ts
Amdoun: this one has been changed

src/dd-draggable.ts
adumesny: | every single mousemove call
the scale/xform should not be called every move event - we know that info when we start/enter/leave - that should be fixed instead.

Amdoun: dragTransform is only calculated in onStartMoving now

src/gridstack.ts
```diff
@@ -2035,11 +2037,29 @@ export class GridStack {
...
       helper = helper || el;
+
+      // if the element is being dragged from outside, scale it down to match the grid's scale
+      // and slightly adjust its position relative to the mouse
+      if (!node.grid?.el) {
+        // this scales the helper down
+        helper.style.transform = `scale(${1 / this.dragTransform.xScale},${1 / this.dragTransform.yScale})`;
+        // this makes it so that the helper is well positioned relative to the mouse after scaling
+        const helperRect = helper.getBoundingClientRect();
+        helper.style.left = helperRect.x + (this.dragTransform.xScale - 1) * (event.clientX - helperRect.x) / this.dragTransform.xScale + 'px';
```
adumesny, line 2048: not sure I understand the "slightly adjust its position" and code here - it s mapping exactly to the same relative point (vido seem that way). can't figure from the code right now

Amdoun: Yes, that code is mapping to the same relative position after scaling. the mouse position relative to the element should be the same after scaling

adumesny:
thanks for making these changes!

src/gridstack.ts
creage: @Amdoun @adumesny These lines (setting `left` and `top` on `helper`) doesn't work well when the draggable has different `appendTo` than `body` (defined in `setupDragIn`), and when this `appendTo` element has `top`/`left` greater than 0.

Helper element receives mouse positions + offset of the `appendTo` element, placing it wrong.

adumesny: @creage, please file a bug with a test case showing the issue. I'll have to remind myself why we have `appendTo:element` to start with (likely historical jquery UI option) as I'm not sure the use case - if you can comment on that as well.
supporting too many options makes the lib complicated/briddle.

creage: @adumesny Well, we don't want to add draggable helper to the body, but we want them to stick under some parent container - this way we can adjust styling of such helpers easily, and less pollute to the global `body` scope.

Like, a Dialog that shows a GridStack with some items we can drag to the grid. All styling is defined within the Dialog component, which we can use in different areas in our app.

So, this option is quite useful - please don't remove it.

adumesny: | All styling is defined within the Dialog component

ok makes sense. but then aren't items clipped by the dialog when dragging around which is a little odd, even if you can only drop them in the grid of the dialog.

PR #2588, opened 2023-12-25
add `on` function overloading for better type hints

liqiuqiui:
### Description
Add type overloading to the `on` and `off` function to obtain better type hints in development. An example is shown in the following figure:

image
image
image
image

### Checklist


adumesny:
thnaks, this handles the string case as well (might need to make an array option as well).

PR #2590, opened 2024-01-02
demo: nested drop

fredericrous:
### Description

show how to add a subgrid with widgets inside either programmatically or by drag and drop. what do you think?

### DIFF
--- a/demo/nested.html
+++ b/demo/nested.html
@@ -16,9 +16,23 @@
       Use v9.2 <b>sizeToContent:true</b> on first subgrid item parent to grow/shrink as needed, while leaving leaf green items unchanged.<br>
       Uses v3.1 API to load the entire nested grid from JSON.<br>
       Nested grids uses v5 <b>column:'auto'</b> to keep items same size during resize.</p>
-    <a class="btn btn-primary" onClick="addNested()" href="#">Add Widget</a>
+    <div style="display: flex; flex-direction: row; gap: 5px;">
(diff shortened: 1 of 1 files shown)


demo/nested.html
```diff
@@ -16,9 +16,23 @@ <h1>Nested grids demo</h1>
...
       Nested grids uses v5 <b>column:'auto'</b> to keep items same size during resize.</p>
-    <a class="btn btn-primary" onClick="addNested()" href="#">Add Widget</a>
+    <div style="display: flex; flex-direction: row; gap: 5px;">
+    <a class="btn btn-primary" onClick="addWidget()" href="#">Add Widget</a>
     <a class="btn btn-primary" onClick="addNewWidget('.sub1')" href="#">Add Widget Grid1</a>
     <a class="btn btn-primary" onClick="addNewWidget('.sub2')" href="#">Add Widget Grid2</a>
+    <a class="btn btn-primary" onClick="addNested()" href="#">Add Widget With Nested Content</a>
+    <div style="width: 100px;">
+      <div
+          class="gridstack-add-widget grid-stack-item ui-draggable"
```
adumesny, line 26: this isn't the right way (and event handler below) to drag a nested grid... but I will commit your change and modify afterward.

adumesny: I've commited my changes this morning to the demo... 'dropped' event is better suited, but you would need the custom attr if you have more than 1 item type to drag in (not case of demo).

PR #2596, opened 2024-01-12
fix(ssr): problem with document not existing on server side

stephenjason89:
### Description
Fixed the problem where document doesn't exist on server side

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -94,6 +94,9 @@ export class GridStack {
    * let grid = document.querySelector('.grid-stack').gridstack;
    */
   public static init(options: GridStackOptions = {}, elOrString: GridStackElement = '.grid-stack'): GridStack {
+    if(typeof document === 'undefined') {
+      return null
+    }
     let el = GridStack.getGridElement(elOrString);
     if (!el) {
       if (typeof elOrString === 'string') {
@@ -121,6 +124,9 @@ export class GridStack {
    */
   public static initAll(options: GridStackOptions = {}, selector = '.grid-stack'): GridStack[] {
     let grids: GridStack[] = [];
+    if(typeof document === 'undefined') {
+      return grids
+    }
     GridStack.getGridElements(selector).forEach(el => {
       if (!el.gridstack) {
         el.gridstack = new GridStack(el, Utils.cloneDeep(options));
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -56,7 +56,7 @@ export class Utils {
   /** convert a potential selector into actual list of html elements. optional root which defaults to document (for shadow dom) */
   static getElements(els: GridStackElement, root: HTMLElement | Document = document): HTMLElement[] {
     if (typeof els === 'string') {
-      const doc = ('getElementById' in root) ? root as Document : undefined;
+      const doc = ('getElementById' in (root || {})) ? root as Document : undefined;
 
       // Note: very common for people use to id='1,2,3' which is only legal as HTML5 id, but not CSS selectors
       // so if we start with a number, assume it's an id and just return that one item...
@@ -66,10 +66,10 @@ export class Utils {
         return el ? [el] : [];
       }
 
-      let list = root.querySelectorAll(els);
+      let list = root?.querySelectorAll(els);
       if (!list.length && els[0] !== '.' && els[0] !== '#') {
-        list = root.querySelectorAll('.' + els);
-        if (!list.length) { list = root.querySelectorAll('#' + els) }
+        list = root?.querySelectorAll('.' + els);
+        if (!list.length) { list = root?.querySelectorAll('#' + els) }
       }
       return Array.from(list) as HTMLElement[];
     }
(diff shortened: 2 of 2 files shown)


src/utils.ts
```diff
@@ -56,7 +56,7 @@ export class Utils {
   /** convert a potential selector into actual list of html elements. optional root which defaults to document (for shadow dom) */
   static getElements(els: GridStackElement, root: HTMLElement | Document = document): HTMLElement[] {
     if (typeof els === 'string') {
```
adumesny, line 58: I would just return if root is null instead of all these checks...

src/gridstack.ts
```diff
@@ -94,6 +94,9 @@ export class GridStack {
    * let grid = document.querySelector('.grid-stack').gridstack;
    */
   public static init(options: GridStackOptions = {}, elOrString: GridStackElement = '.grid-stack'): GridStack {
+    if(typeof document === 'undefined') {
```
adumesny, line 97: I don't understand how returning null and doing nothing on the server side is going to help anything - not only will the dom not be created with correct html, but the caller will likely crash refering the grid.

src/utils.ts
stephenjason89: Do you mean return undefined since root is undefined on server side? 

I'll change it to that.

src/gridstack.ts
stephenjason89: We want it to not do anything in the server side, like accessing document because this is undefined there. Since DOM is not present there, we cant manipulate the DOM during SSR.

fredericrous:
I tested this PR, the warnings indeed are removed thanks. I think this PR is a first step toward ssr compatibility

Of course in an ideal world, supporting srr would mean that gridstack serves a prebuilt grid, created server side.

stephenjason89:
@fredericrous thank you for testing it out. Indeed, a prebuilt grid would be nice. I hope to see it soon as well 😊

adumesny:
| Of course in an ideal world, supporting srr would mean that gridstack serves a prebuilt grid, created server side

my point exactly - this feels like a bandaid. the all point of SSR is that all the HTML (and hence the doms attr, style and **widget content** which can be created when loading a grid definition (eg angular components) are rendered and sent to the user before all JS code is sent over.

And why would init() be calledin the server (no-op) AND client side (actually do something) ? everything runs twice then and need to handle already done stuff (which GS does I beleive).

src/gridstack.ts
adumesny: yes if not needed - skipping is likely wrong anyway.

src/utils.ts
adumesny: yes if not needed, or better fix as I suggested.

adumesny:
no my worry is that init/load can do a lot more work than just setting dom attributes. in dynamic cases it causes the entire widget and content to be created dynamically. so you would miss the entire page html rendering ahead of time, which is the all point of SSR. eg https://github.com/gridstack/gridstack.js/blob/master/angular/projects/demo/src/app/app.component.ts#L41

fredericrous:
I'm not sure I understand your point. Discussing this on a forum might not be the best approach.

My perspective: we determine what to serve based on the presence of window.document. For server-side rendering (SSR), we can't use document.querySelector or addEventListener.

When window.document is available, we attach event listeners and use our preferred framework to manage the content within widgets. For example, I use React to replace the content of a widget by identifying its ID, rather than using the widget's "content" method. On the other hand, in SSR where window.document is absent, we could use the renderToString function from our respective frameworks (in my case, I'd use the "content" method to inject html from renderToString).

GridStack heavily relies on querySelectorAll, for instance, to identify all subgrids. In SSR, this could be substituted with a recursive function that iterates over the GridStack instance, checking for gridstackNode.subGrid and recursively calling itself. There could be more efficient solutions to reduce time complexity

Regarding helping on SSR feature, at this time I can only offer to do reviews. I need to get my MVP working, I have no revenue yet. And SSR will be a need if my project works but at this time it is a nice to have
In the mid-time, Stephen's PR prevents frameworks like Remix to crash. It's not a hard crash but it does spam the console heavily. I would merge it once the fix you asked are implemented.

stephenjason89:
> my point exactly - this feels like a bandaid. the all point of SSR is that all the HTML (and hence the doms attr, style and **widget content** which can be created when loading a grid definition (eg angular components) are rendered and sent to the user before all JS code is sent over.

Yes this is exactly what it is, to just avoid all the SSR errors when gridstack try to access DOM during SSR. 
I look forward to an actual implementation on SSR where a grid can be initially created, I do not have the time to implement it now.

stephenjason89:
> my point exactly - this feels like a bandaid. the all point of SSR is that all the HTML (and hence the doms attr, style and **widget content** which can be created when loading a grid definition (eg angular components) are rendered and sent to the user before all JS code is sent over.

Absolutely, you've hit the nail on the head. This is essentially a temporary fix to bypass the errors that occur when gridstack attempts to access the DOM during SSR.

I'm eager to see a genuine implementation for SSR where a grid can be set up initially. Currently, I lack the time to work on this myself.

I'm looking forward to this PR being merged promptly to prevent SSR errors. Ideally, having comprehensive SSR support would be highly advantageous.

src/utils.ts
stephenjason89: done, i have reverted the file.

stephenjason89:
> my point exactly - this feels like a bandaid. the all point of SSR is that all the HTML (and hence the doms attr, style and **widget content** which can be created when loading a grid definition (eg angular components) are rendered and sent to the user before all JS code is sent over.

Indeed, this is precisely the case. The primary aim is to circumvent the errors that occur in SSR when Gridstack attempts to access the DOM. Although I eagerly anticipate a genuine SSR implementation where a grid can be initially set up, I currently lack the time to undertake this task.

I'm looking forward to the prompt merging of this PR, as it will help us avoid SSR-related errors. However, having complete support for SSR would be highly beneficial and is greatly desired.

PR #2612, opened 2024-02-04
restrict vertical resize with sizeToContent

adumesny:
### Description
* since sizeToContent:true will automatically size the content vertically, don't let the user change the height to start with

Note: tried to change the feedback to have e,w arrows but we only have 10px area which is too narrow to show feedback (we have 20x20 for corners that may overlap content which is ok). Showing normal handles for now.

### Checklist

### DIFF
--- a/src/dd-gridstack.ts
+++ b/src/dd-gridstack.ts
@@ -40,9 +40,18 @@ export class DDGridStack {
       } else if (opts === 'option') {
         dEl.setupResizable({ [key]: value });
       } else {
-        const grid = dEl.el.gridstackNode.grid;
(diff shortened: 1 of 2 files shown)


cff3:
Hi @adumesny as far as I can see, this change prevents a user from resizing a widget vertically even if `sizeToContent` was set to a number instead of setting it to `true`. Is this intentional? 

On the demo page (https://gridstackjs.com/demo/sizeToContent.html) it is still possible to resize widget E, but maybe the demo site uses an older version of grid stack? I didn't manage to understand what version is used on the demo site.

adumesny:
It's a bug. Number should let you resize that's the all point. I made the tweak too fast...

yusufkandemir:
This is a breaking change. I am yet to understand if `sizeToContent: number` in nodes or anything else is an alternative viable solution to preserve the old behavior. But, the migration guide should definitely mention this change, preferably with an alternative.

PR #2617, opened 2024-02-16
empty addedNodes and removedNodes before triggering event to prevent infinite loop

galizhur:
If you call addWidget inside the grid added or removed events, the items passed in the event for the 

Calling **addWidget** inside the **"added"** event triggers another **"added"** event with the **items** argument containing the newly added item and the initial item that triggered the first **added** event. The second **"added"** event should contain only the newly added item.

This happens because the **"added"** event is triggerd first and the internal **addedNodes** array is emptied afterwards. The fix was to empty the **addedNodes** array before triggering the **"added"** event.

The same happens when calling **removeWidget** inside the **"removed"** event.

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -7,9 +7,11 @@
  */
 import { GridStackEngine } from './gridstack-engine';
 import { Utils, HeightData, obsolete, DragTransform } from './utils';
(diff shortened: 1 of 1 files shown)


adumesny:
would be great if you didn't reformat/pretty the existing code...

galizhur:
> would be great if you didn't reformat/pretty the existing code...

yeah. sorry about that. i didn't notice there were so many reformats :(

PR #2626, opened 2024-03-06
Synchronize to base repository

manuel-rw:


adumesny:
@Meierschlumpf @manuel-rw this is lame. rather than contribute andwork within the lib to make it work better for React you just fork it off and add an odd flag to hide items - I suspect you expect React to destroy and re-create the item in the other grid.
I could have done that for Angular as well, but moving the item and fixing the framework to know it moved and not recreate everything (which could be expensive) is a better flow.

Also you are selling a product based on GS, yet don't bother donating or even contributing back... downside of doing open source - people just take advantage of you.

rant off...

Meierschlumpf:
Just for your info, we're an open source project as well and so we just use your great library. We wanted to make it work, but we're not able to find a better solution than that which we used for our project with that flag. I've created a pull request with the exact changes that I made in our fork and you didn't want it. Thanks again for the good library...

manuel-rw:
Hi, I'd like to chime in as well.
We do not sell any product and all of our code is open source. I don't know what product you're referring to but I did make a donation a while back.
We work on this project at a loss without any profit if monetary gains from your code are a concern.

We had specific requirements that didn't fully work with your library so we decided to make this fork.
We don't use Angular so that's why it was removed - why maintain code that we don't use?
My intention with this pull request was to synchronize changes to our fork.
GitHub just switched the target while I was creating the PR.

Our goal is not to anger you or any contributors.
We have asked in the past for better support for React and the interest was a bit low - hence the fork.

It's normal that the opinions of many developers don't always align - that's also something good.

Up to you what you want to do with this information / pull request. I certainly did not wish to anger you and I wish you a good evening / rest of your day.

Cheers 🥂

adumesny:
sorry if I jumped the gun.

| I've created a pull request with the exact changes that I made in our fork and you didn't want it
I recall seeing it now. I don't know React, so I've asked for help getting a similar wrapper than the angular one I ship. I just know that doing the *ngFor loop (where Angular does the actual widget creation list) is problematic as we also want GS to add and re-parent widgets are part of drag&Drop, and synching the 2 are problematic, which is why I took the approach I did - I assume the same issue can be solved in React -> let React create components (given a parent grid) as needed and unref them when told.

I also assume you just temporarily hide a removed widget to let React redo the dom updates into new location instead ?
yeah I could have done that as well, but not ideal (re-create).

manuel-rw:
I just wanted to let you know that we are both currently busy with our private projects (+ surgery) and are unable to contribute at the moment.
Would you be willing to accept a contribution regarding this in the future?
We are not willing to invest time into this again for it to be rejected or leading to nothing (as it was the case when we wanted to do this in the past).

Looking forward to your reply. I cannot answer most of your questions because Meier did the majority of the work with the library.

adumesny:
I don't want a hack in the library - need to understand why React can't be told a widget moved and act accordingly - that's what I do for Angular.

manuel-rw:
@Meierschlumpf should know the answer for that.

Meierschlumpf:
I think the issue with react is the reactivity which updates the virtual DOM and after that wants to remove no longer rendered elements. I had no issue using gridstack and react together until I've added a second stack so items can be moved from one to the other. Because I needed to update the location and parent, it rerendered and wanted to remove the item from the previous parent, which did not exist anymore because it was already removed by gridstack.

PR #2630, opened 2024-03-10
fix/2584: refactored sort function

### LINKED ISSUE #2584: [BUG] The order of the grids can break, when the page is loading on small screen, with Responsive Layouts
## Subject of the issue
What the title says. 
The order of the grids can break, when the page is loading on small screen, with Responsive Layouts.

## Your environment
version of gridstack.js - 10.0.1
which browser/OS - Chrome/MacOS

## Steps to reproduce

https://gridstackjs.com/demo/responsive_break.html

In the Responsive demos (either by column size or using breakpoints) if you reload the page when only one column should be displayed, than grid nr. 5 is placed before grid nr. 4. (But if you start from a bigger resolution, and zoom to the smallest, the order remains fine).

## Expected behavior
The order should remain 1,2,3,4,5.

The bug is even more visible, if you change the items array a bit.
let items = [ 
      {x:0,y:0,w:1,h:1},
	  {x:1,y:0,w:1,h:1},
	  {x:2,y:0,w:1,h:2},
	  {x:9,y:0,w:1,h:2},
	  {x:0,y:1,w:1,h:1},
	  {x:1,y:1,w:1,h:1},
	  {x:1,y:2,w:2,h:2}
    ];

JakubEleniuk:
### Description
Fixes #2584

From my manual testing everything looks fine.

### Checklist

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -146,14 +146,15 @@ export class Utils {
    * Sorts array of nodes
    * @param nodes array to sort
    * @param dir 1 for asc, -1 for desc (optional)
-   * @param width width of the grid. If undefined the width will be calculated automatically (optional).
+   * @param column number of columns in the grid. If undefined columns will be calculated automatically (optional).
    **/
   static sort(nodes: GridStackNode[], dir: 1 | -1 = 1, column?: number): GridStackNode[] {
     column = column || nodes.reduce((col, n) => Math.max(n.x + n.w, col), 0) || 12;
-    if (dir === -1)
-      return nodes.sort((a, b) => ((b.x ?? 1000) + (b.y ?? 1000) * column)-((a.x ?? 1000) + (a.y ?? 1000) * column));
-    else
-      return nodes.sort((b, a) => ((b.x ?? 1000) + (b.y ?? 1000) * column)-((a.x ?? 1000) + (a.y ?? 1000) * column));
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -146,14 +146,15 @@ export class Utils {
...
+   * @param column number of columns in the grid. If undefined columns will be calculated automatically (optional).
    **/
   static sort(nodes: GridStackNode[], dir: 1 | -1 = 1, column?: number): GridStackNode[] {
     column = column || nodes.reduce((col, n) => Math.max(n.x + n.w, col), 0) || 12;
-    if (dir === -1)
-      return nodes.sort((a, b) => ((b.x ?? 1000) + (b.y ?? 1000) * column)-((a.x ?? 1000) + (a.y ?? 1000) * column));
-    else
-      return nodes.sort((b, a) => ((b.x ?? 1000) + (b.y ?? 1000) * column)-((a.x ?? 1000) + (a.y ?? 1000) * column));
+    return nodes.sort((a, b) => {
+      let diffY = dir * (a.y - b.y);
```
adumesny, line 154: this is incorrect becaue x and y could be undefined (auto position) which is why I default to 1000 to be last, but still sorted relative to other items.
But looking at your code I realize I don't really need column at all since it's just relative and Y trumps anything else...

Issue #2667, opened 2024-04-24
Dragging toolbar item in + out + in doesn't honour gs-w attribute sizes


BussiGiu:
## Subject of the issue
I have a toolbar with one widget and attributes set to `gs-w="1"` and `gs-h ="1"`. 
When I try to drag and drop the widget inside the grid it usually works fine, but if I accidentally first drag it inside and outside of the grid and then drop it inside, the widget ends up occupying a space that doesn't honour the width and height attributes. 
Also, the attributes set into the HTML disappear and from now on, they are completely ignored.

I'm using Gridstack in Angular and the widget toolbar implements `BaseWidget` class but I also managed to reproduce the issue without Angular, so I don't think Angular makes any difference here.

## Your environment
* gridstack.js - 10.1.2
* angular - 16.1.2 (not necessary to reproduce)

## Steps to reproduce
I recreated the bug in this demo: https://jsfiddle.net/gusn6cvy/11/
And here's is a video of the effect:

https://github.com/gridstack/gridstack.js/assets/46027864/2b3ba3c3-65e5-4fee-a901-db9f2c2babe3

## Expected behaviour
Dragging a widget from a toolbar should create a widget on the grid that maintains the original attributes (`gs-w` and `gs-h`) even if it's dragged inside and outside of the grid many times.

dinozof:
I'm having the same issue as well!

adumesny:
thanks for the report. When leaving the grid we try to find if it came from another grid and restore that but this appears to mess the toolbar...

AndreasHerss:
@BussiGiu i ran into the same issue as you, first time it gives 1x1 next time a 2x1. 

My solution was to change the setupDragIn to contain definitions for the widget like so
`GridStack.setupDragIn('.newWidget', { appendTo: 'body', helper: 'clone' }, [{w:1, h:1}]);`
If you're having multiple widgets that need different parameters, you can define them in the array too, or just call the setupDragIn multiple times with different dragIn classes and the [{w:1, h:1}] applied :)

adumesny:
@AndreasHerss oh yeah, forgot about this bug. I totally revamped how drag&drop from toolbar works in V11 and added the ability to have WidgetItem assigned to toolbar items, given you full power (since gs- attributes are limited in choices).
[...]

PR #2672, opened 2024-05-02
fix issue on failed drag makes the whole grid unresponsive

mustafamilyas:
### Description
Screenshot 2024-05-02 at 09 16 45
when we drag a widget and fail, it will throw the above error and make all the whole grid unresponsive

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -2251,8 +2251,8 @@ export class GridStack {
         if (wasAdded) {
           this.engine.cleanupNode(node); // removes all internal _xyz values
           node.grid = this;
+          delete node.grid._isTemp;
         }
-        delete node.grid._isTemp;
         dd.off(el, 'drag');
         // if we made a copy ('helper' which is temp) of the original node then insert a copy, else we move the original node (#1102)
         // as the helper will be nuked by jquery-ui otherwise. TODO: update old code path
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -111,9 +111,9 @@ export class Utils {
 
   /** true if we should resize to content. strict=true when only 'sizeToContent:true' and not a number which lets user adjust */
   static shouldSizeToContent(n: GridStackNode | undefined, strict = false): boolean {
-    return n?.grid && (strict ? 
-    (n.sizeToContent === true || (n.grid.opts.sizeToContent === true && n.sizeToContent === undefined)) :
-    (!!n.sizeToContent || (n.grid.opts.sizeToContent && n.sizeToContent !== false)));
+    return n?.grid && (strict ?
+      (n.sizeToContent === true || (n.grid.opts.sizeToContent === true && n.sizeToContent === undefined)) :
+      (!!n.sizeToContent || (n.grid.opts.sizeToContent && n.sizeToContent !== false)));
   }
 
   /** returns true if a and b overlap */


src/utils.ts
```diff
@@ -111,9 +111,9 @@ export class Utils {
 
   /** true if we should resize to content. strict=true when only 'sizeToContent:true' and not a number which lets user adjust */
   static shouldSizeToContent(n: GridStackNode | undefined, strict = false): boolean {
-    return n?.grid && (strict ? 
-    (n.sizeToContent === true || (n.grid.opts.sizeToContent === true && n.sizeToContent === undefined)) :
-    (!!n.sizeToContent || (n.grid.opts.sizeToContent && n.sizeToContent !== false)));
+    return n?.grid && (strict ?
+      (n.sizeToContent === true || (n.grid.opts.sizeToContent === true && n.sizeToContent === undefined)) :
+      (!!n.sizeToContent || (n.grid.opts.sizeToContent && n.sizeToContent !== false)));
```
mustafamilyas, line 116: when I run the test script, it fails, and this is the result when I run `yarn test --fix`

adumesny: that's just reformatting (yarn lint --fix) not yarn test.... humm.

ilyaspalinor: Apologies for any confusion, I realize I forgot to include instructions on how to reproduce the issue. However, I want to clarify that this is not the source of the bug; it's related to other changes. Thank you for your understanding!

src/gridstack.ts
```diff
@@ -2251,8 +2251,8 @@ export class GridStack {
         if (wasAdded) {
           this.engine.cleanupNode(node); // removes all internal _xyz values
           node.grid = this;
+          delete node.grid._isTemp;
         }
-        delete node.grid._isTemp;
```
ilyaspalinor, line 2255: To reproduce the issue:

1. Enable dragging a widget from outside onto the grid.
2. Ensure the grid is crowded, leaving very little space (less than the size of the widget).
3. Drag the widget into that small space.

This will cause the addition of the widget to fail and become unresponsive. If you need further assistance, feel free to reach out. Thank you!

src/utils.ts
mustafamilyas: Apologies for any confusion, I realize I forgot to include instructions on how to reproduce the issue. However, I want to clarify that this is not the source of the bug; it's related to other changes. Thank you for your understanding!

src/gridstack.ts
mustafamilyas, line 2255: To reproduce the issue:

1. Enable dragging a widget from outside onto the grid.
2. Ensure the grid is crowded, leaving very little space (less than the size of the widget).
3. Drag the widget into that small space.

This will cause the addition of the widget to fail and become unresponsive. If you need further assistance, feel free to reach out. Thank you!

adumesny: so this doesn't fix dropping into a full grid ? what does it fix then ?

mustafamilyas: These changes should resolve the issue. The step provided is intended to demonstrate how to reproduce the issue before implementing these changes. I apologize if the explanation is unclear.

mustafamilyas: The issue arises because 'node.grid' is populated only when the widget is added (line 2253). Therefore, if the widget fails to be added, an error occurs because the system attempts to delete 'node.grid._isTemp', which does not exist.

I'll aim to create a reproduction of the issue sometime next week, if that works for you.

adumesny: if that's the case delete node.grid?._isTemp; woudl have suffice. I'll need to debug this to see the implication of move it out...

sfragis:
Hi @mustafamilyas and @adumesny, does this PR also fix issue #2667 by any chance?

adumesny:
@sfragis nothing to do with it.

mustafamilyas:
@adumesny, apologies for the delay. I've made the adjustments as per your suggestion and tested them on my end. It seems to have resolved the issue at least on my side.

Below, you'll find a JSFiddle and a video demonstrating how to reproduce the error before the changes were made. Let me know if you need further assistance. Thank you!

https://jsfiddle.net/redmaze/jqw25fed/7/
https://www.loom.com/share/c89f4254d3e347d4adcc934d6bd71d74?sid=bc050e16-02a4-4170-aa60-c8b70d9936b1

src/gridstack.ts
adumesny: | The issue arises because 'node.grid' is populated only when the widget is added (line 2253)

no it could come from another grid as well... anyway the optional null check is safe and right thing to do, but I may need to exit sooner and not do some work, but for now that fixes the js error. thank you for the demo...

adumesny:
that demo is weird. items are places 1x1 first, then 2x1 (to match actual dragged size, what it should be) even if you don't resize the dropped item.

PR #2675, opened 2024-05-14
fix(save): don't remove w or h

brian-patrick-3:
### Description
GridStack freezes the browser when trying to call `.load()` with one of these values missing. This prevents instances where `.save()` drops a value ex `h`, and subsequently calling `.load()` freezes the browser.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -326,8 +326,6 @@ export class Utils {
     if (!n.noResize) delete n.noResize;
     if (!n.noMove) delete n.noMove;
     if (!n.locked) delete n.locked;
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -326,8 +326,6 @@ export class Utils {
     if (!n.noResize) delete n.noResize;
     if (!n.noMove) delete n.noMove;
     if (!n.locked) delete n.locked;
-    if (n.w === 1 || n.w === n.minW) delete n.w;
```
adumesny, line 329: that is not the right fix. data can stil come with fields missing. the freeze part should be fixed instead

Issue #2695, opened 2024-06-06
Can't cancel an external item from being dragged by pressing 'Esc' key


adrianbenjuya:
## Subject of the issue
As a user I want to cancel Drag&Drop of a new item from outside the grid by pressing 'Esc' key while the item is being dragged. This works fine for items that are already inside the grid.

## Your environment
* Gridstack demo page
* Google Chrome

## Steps to reproduce
Just drag an external item and press 'Esc' key while dragging

![chrome_CTznjhRBKu](https://github.com/gridstack/gridstack.js/assets/17908930/655452b8-6a0d-4136-bf2a-5ff35295503e)

## Expected behavior
Dragging action should be cancelled, or if not possible at least handle the error.

adumesny:
I added Esc key as an addition to a user paying for the 'r' feature, but I'm not planning to add more options to it (like dragging from outside, which yeah ideally would work too) unless someone want to donate for it.

adrianbenjuya:
Is it fine if I try to fix it and create a PR?

adumesny:
fixed in next release. don't forget to donate if you find this lib useful!

PR #2698, opened 2024-06-13
Bugfix: prevent throwing error when node.grid is undefined (drag item from outside the grid)

adrianbenjuya:
### Description
I added an extra nullcheck before performing the key event callback. If `n` or `n.grid` aren't defined then just `return`.

### Checklist


adumesny:
thanks. the real fix would handle null grid (dragginbg from outside) case for #2695

PR #2709, opened 2024-06-23
support for multiple drag handles

nickfulcher:
### Description
Currently when options handle is defined it applies event listeners to the first instance that has the handle class. These changes add the events to all elements that has the handle class allowing for multiple drag handles with in a grid item

### DIFF
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -55,6 +55,8 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   protected dragElementOriginStyle: Array<string>;
   /** @internal */
   protected dragEl: HTMLElement;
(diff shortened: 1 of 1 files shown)


src/dd-draggable.ts
```diff
@@ -55,6 +55,8 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   protected dragElementOriginStyle: Array<string>;
   /** @internal */
   protected dragEl: HTMLElement;
```
adumesny, line 57: you should remove that if you are using a list instead... and fix line  146

adumesny:
fixed your incorrect code... didn't get to test touch devices, hopefully didn't break e.target case....

PR #2848, opened 2024-11-05
feat: updated react wrapper to support for V11 along with section support

GowthamTG:
### Description
This PR adds support for gridstack going forward with gridstack V11 using the `renderCB` function and this example also adds support of sections in react

https://github.com/user-attachments/assets/bcbc522d-09e6-4d98-a6ea-1ae2020a360e

### Checklist


adumesny:
funny you added 'section' nested grid widget, which is exactly what I'm doing at work, hence my changes to v11.
you should use the demo css styles so it looks like the nested.html example for consistency and easier to see nesting...

image

react/lib/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+  { c: 6, w: 950 },
+  { c: 8, w: 1100 },
+];
+
+export const SUB_GRID_OPTIONS: GridStackOptions = {
+  acceptWidgets: true,
+  cellHeight: 54,
+  column: 'auto',
+  columnOpts: {
+    breakpoints: BREAKPOINTS,
```
adumesny, line 15: nested grid either use column:'auto' (get size from parent) OR breakpoint, not both... I use auto so dragging in/out keeps items the same which is more common.

react/lib/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+export const BREAKPOINTS = [
+  { c: 1, w: 700 },
+  { c: 3, w: 850 },
+  { c: 6, w: 950 },
+  { c: 8, w: 1100 },
+];
+
+export const SUB_GRID_OPTIONS: GridStackOptions = {
+  acceptWidgets: true,
+  cellHeight: 54,
```
adumesny, line 12: odd number :)

react/lib/gridstack-context.tsx
```diff
@@ -1,61 +1,91 @@
...
+      GridStack.renderCB = renderCBFn;
+      return GridStack.init(optionsRef.current, containerRef.current);
+    }
+    return null;
+  }, [renderCBFn]);
+
+  useLayoutEffect(() => {
+    if (!isEqual(options, optionsRef.current) && parentGrid) {
+      try {
+        parentGrid.removeAll(false);
```
adumesny, line 56: this are some of the weird part about React and GS not being in sync and why in my angular wrapper I do NOT create the grid or items in html template but let GS call me back t create the appropriate component (based on NgWidgetItem.selector)

react/package-lock.json
```diff
@@ -0,0 +1,2061 @@
+{
```
adumesny, line 1: what is that file vs yarn.lock, which should not have changed much from you just adding 1 new package...

react/yarn.lock
```diff
@@ -2,136 +2,26 @@
...
-
-"@esbuild/android-x64@0.21.5":
-  version "0.21.5"
-  resolved "https://registry.yarnpkg.com/@esbuild/android-x64/-/android-x64-0.21.5.tgz#29918ec2db754cedcb6c1b04de8cd6547af6461e"
-  integrity sha512-D7aPRUUNHRBwHxzxRvp856rjUHRFW1SdQATKXH2hqA0kAZb1hKmi02OpYRacl0TxIGz/ZmXWlbZgjwWYaCakTA==
-
 "@esbuild/darwin-arm64@0.21.5":
   version "0.21.5"
-  resolved "https://registry.yarnpkg.com/@esbuild/darwin-arm64/-/darwin-arm64-0.21.5.tgz#e495b539660e51690f3928af50a76fb0a6ccff2a"
+  resolved "https://registry.npmjs.org/@esbuild/darwin-arm64/-/darwin-arm64-0.21.5.tgz"
```
adumesny, line 7: please leave it using yarnpkg.com like all other ones.

adumesny (commented):
that is great! thnak you for updating the React code. left some comments to address though.

react/lib/gridstack-context.tsx
GowthamTG: This is for when the top options change we are destroying the current grid and reinitialising the same 
one of the thing why we are not letting GS dom manipulation is that React has their own dom manipulation logic 
so if we are interfering with the same things are not working as expected because as the grid element is already removed by react the the internal GS does not expect the same to happen

react/lib/constants.ts
adumesny: humm, again I would do `column:'auto'` for nested. you happen to have a section take entire width of the grid so you are lucking out that parent & child matches, but this would break if it wasn't full width.
we can update later...

react/src/demo/demo.css
```diff
@@ -1,17 +1,77 @@
...
+.sidebar > .grid-stack-item,
+.grid-stack-item-content {
+  text-align: center;
+  background-color: #18bc9c;
+}
+
+.grid-stack-item-removing {
+  opacity: 0.5;
+}
+.trash {
```
adumesny, line 53: this can all be trimmed out... or re-use the main demo file by reference instead...

adumesny (commented):
great thanks. some items left for next CL.

PR #2850, opened 2024-11-13
- Fixed test compilation

lmartorella:
### Description

The `test` target is showing a lot of Typescript compilation errors.
Some of them are due to missing "include" filter in the karma TS section (and so karma was ending up in trying to compile samples, etc..).
However other TS errors in spec files seems being due to genuine API breaks.

With this PR the compiler phase is now working.

However it seems that there are a lot of dangling errors in automated tests. I cannot see any issue or PR opened for this. Is `npm run test` / `yarn test` working on your side?

```
Chrome Headless 131.0.0.0 (Mac OS 10.15.7): Executed 62 of 171 (12 FAILED) DISCONNECTED (31.872 secs / 0.038 secs)
```

Thx, L.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,9 +15,6 @@
     "strict": false,
     "target": "ES2020"
   }, 
-  "exclude": [
-    "./src/**/*.spec.ts",
-  ],
   "include": [
     "./src/**/*.ts"
   ],
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -14,7 +14,10 @@ module.exports = function(config) {
       //     }
       //   }
       // },
-      exclude: ["demo", "dist/ng"] // ignore dummy demo .ts files
+      exclude: ["demo", "dist/ng"], // ignore dummy demo .ts files
+      include: [
+        "./spec/**/*-spec.ts"
+      ]
     },
 
     // base path that will be used to resolve all patterns (eg. files, exclude)
--- a/spec/gridstack-engine-spec.ts
+++ b/spec/gridstack-engine-spec.ts
@@ -146,21 +146,21 @@ describe('gridstack engine', function() {
     beforeAll(function() {
       engine = new GridStackEngine({float:true});
       engine.nodes = [
-        engine.prepareNode({x: 0, y: 0, id: 1, _dirty: true}),
-        engine.prepareNode({x: 3, y: 2, w: 3, h: 2, id: 2, _dirty: true}),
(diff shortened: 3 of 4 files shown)


lmartorella:
Fixed the DISCONNECTED issue as well, due to infinite loop in the fix collision code. Added a guard to quit the loop in case of probably infinite loop.
In order to keep the guard as simple and performant as possible, the exit check is based on counter, compared to the magnitude of nodes contained.

adumesny:
hey thanks for looking at restoring `yarn test` back, it stopped working when I did some big collision overall and didn't spend the time to fix what could be a valid error or incorrect old behavior (the latter I suspect). then bunch of folders (like React) was added and things don't run.

tsconfig.json
```diff
@@ -15,9 +15,6 @@
     "strict": false,
     "target": "ES2020"
   }, 
-  "exclude": [
```
adumesny, line 18: that may force .spec.ts files to be in the npm build package which we don't want...

src/gridstack-engine.ts
```diff
@@ -100,7 +100,11 @@ export class GridStackEngine {
 
     let didMove = false;
     const newOpt: GridStackMoveOpts = {nested: true, pack: false};
+    let counter = 0;
     while (collide = collide || this.collide(node, area, opt.skip)) { // could collide with more than 1 item... so repeat for each
+      if (counter++ > this.nodes.length * 2) {
```
adumesny, line 105: I need to see case where this happens before taking this workaround... will revert it locally.

tsconfig.json
lmartorella: There are no .spec files in ./src, so this is safe to delete.

src/gridstack-engine.ts
lmartorella: This is actually fixing a broken test, so it is easy to replicate:

> Chrome Headless 131.0.0.0: gridstack grid.min/max width/height should set gs-min-w to 2. FAILED
> Error: Infinite collide check

Without this safe check, the test suite is simply going in infinite loop causing "DISCONNECTED" on headless or a stuck browser when testing on real browser. In both cases, no results are produced:

>  Disconnected reconnect failed before timeout of 2000ms (ping timeout)

adumesny: FYI, I fixed the infinite loop (had NaN for maxW) but left the check in there in case...

PR #2851, opened 2024-11-13
Default columns arg for non-12 columns scenarios

lmartorella:
### Description

Some applications can use default column number (for the largest resolutions) higher than 12. 
In that case, during column count dynamic switch due to responsive-based resizes, it can happen that the `nodeBoundFix` function would try to cap the coordinates to 12 rather than the default max column number.
With this fix it is then possible to subclass `GridStackEngine` and change the default "max" column, rather having the constants hard-coded in the codebase.

Thx, L.

### Checklist
- [x] All NEW tests passing (`yarn test`) (unfortunately the main trunk is not stable at the moment)

### DIFF
--- a/spec/gridstack-engine-spec.ts
+++ b/spec/gridstack-engine-spec.ts
@@ -371,6 +371,53 @@ describe('gridstack engine', function() {
     });
   });
   
+  describe('test columnChanged and save', function() {
+    beforeAll(function() {
+    });
+    it('wont\'t break layouts with 12 columns', function() {
+      engine = new GridStackEngine({ column: 12 });
+      // Add two side-by-side components 6+6 = 12 columns
+      engine.addNode({ x: 0, y: 0, w: 6, h: 1, id: "left" });
+      engine.addNode({ x: 6, y: 0, w: 6, h: 1, id: "right" });
+      engine.save().forEach(node => engine.nodeBoundFix(findNode(engine, node.id!)!));
+      expect(findNode(engine, "left")).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 6, h: 1}));
+      expect(findNode(engine, "right")).toEqual(jasmine.objectContaining({x: 6, y: 0, w: 6, h: 1}));
+      // Resize to 1 column
+      engine.column = 1;
+      engine.columnChanged(12, 1);
+      engine.save().forEach(node => engine.nodeBoundFix(findNode(engine, node.id!)!));
+      expect(findNode(engine, "left")).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 1, h: 1}));
+      expect(findNode(engine, "right")).toEqual(jasmine.objectContaining({x: 0, y: 1, w: 1, h: 1}));
+      // Resize back to 12 column
+      engine.column = 12;
+      engine.columnChanged(1, 12);
+      engine.save().forEach(node => engine.nodeBoundFix(findNode(engine, node.id!)!));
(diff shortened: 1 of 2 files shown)


src/gridstack-engine.ts
```diff
@@ -31,6 +31,7 @@ export class GridStackEngine {
   public addedNodes: GridStackNode[] = [];
   public removedNodes: GridStackNode[] = [];
   public batchMode: boolean;
+  public defaultColumn = 12;
```
adumesny, line 34: I think instead of having a 'hard coded' max layout of 12, we can just use the largest size in the layout as the current max. will have to check code again...

adumesny:
| With this fix it is then possible to subclass GridStackEngine and change the default "max" column, rather having the constants hard-coded in the codebase.

we shouldn't need to subclass to set the 'max column' layout saving. will add an option of this instead...

spec/gridstack-engine-spec.ts
```diff
@@ -371,6 +371,53 @@ describe('gridstack engine', function() {
...
+      expect(findNode(engine, "left")).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 12, h: 1}));
+      expect(findNode(engine, "right")).toEqual(jasmine.objectContaining({x: 12, y: 0, w: 12, h: 1}));
+      // Resize to 1 column
+      engine.column = 1;
+      engine.columnChanged(24, 1);
+      engine.save().forEach(node => engine.nodeBoundFix(findNode(engine, node.id!)!));
+      expect(findNode(engine, "left")).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 1, h: 1}));
+      expect(findNode(engine, "right")).toEqual(jasmine.objectContaining({x: 0, y: 1, w: 1, h: 1}));
+      // Resize back to 24 column
+      engine.column = 24;
```
adumesny, line 413: note that this would work BEFORE your changes too as you are calling engine.columnChanged() directly with 24 and defaultColumn is not updated to match

adumesny:
@lmartorella this works so not sure we needed this code after all... we save to 12 column the 24 layout info and reuse it anyway... maybe some corner case with scaling would be affected, not sure...

```ts
    it('24 layout in 12 column to 1 back 12 then 24', function() {
      const children: GridStackWidget[] = [{ x: 0, y: 0, w: 12, h: 1, id: 'left' }, { x: 12, y: 0, w: 12, h: 1, id: 'right' }];
      children.forEach((c, i) => c.content = c.id);

      grid = GridStack.init({children});
      const left = find('left');
      const right = find('right');

      // side-by-side components 12+12 = 24 columns but only have 12!
      expect(left).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 12}));
      expect(right).toEqual(jasmine.objectContaining({x: 0, y: 1, w: 12}));
      // Resize to 1 column
      grid.column(1);
      expect(left).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 1}));
      expect(right).toEqual(jasmine.objectContaining({x: 0, y: 1, w: 1}));
      // Resize back to 12 column
      grid.column(12);
      expect(left).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 12}));
      expect(right).toEqual(jasmine.objectContaining({x: 0, y: 1, w: 12}));
      // Resize to 24 column
      grid.column(24);
      expect(left).toEqual(jasmine.objectContaining({x: 0, y: 0, w: 12}));
      expect(right).toEqual(jasmine.objectContaining({x: 12, y: 0, w: 12}));
    });
```

PR #2854, opened 2024-11-14
Removed dynamic stylesheet and migrated to CSS vars

lmartorella:
### Description

The proposed change is to remove the dynamic CSS classes generation for rows and margins.
This has many benefits: it will speed up relayout operations, optimize performances for high row count and and fix the CSP style compatibility.

The requirement is CSS variables, that is now widely available: https://caniuse.com/css-variables

This deprecates the `styleInHead` properties as well.

It overrides #2852 as well.

### Checklist
- [ ] All tests passing (`yarn test`) (well, the failing test count is not increasing after the change, the main trunk has many failures)

### DIFF
--- a/spec/gridstack-spec.ts
+++ b/spec/gridstack-spec.ts
@@ -1361,35 +1361,7 @@ describe('gridstack', function() {
       let grid = GridStack.init(options);
       expect(grid.el.classList.contains('grid-stack-rtl')).toBe(false);
     });
-  });
-  
-  describe('grid.opts.styleInHead', function() {
-    beforeEach(function() {
-      document.body.insertAdjacentHTML('afterbegin', gridstackHTML);
-    });
-    afterEach(function() {
-      document.body.removeChild(document.getElementById('gs-cont'));
-    });
-    it('should not add STYLE to parent node', function() {
-      const options = {
-        cellHeight: 80,
-        verticalMargin: 10,
-        float: false,
-      };
-      GridStack.init(options);
-      expect(document.querySelector("style")).toBe(null);
-    });
-    it('should not add STYLE to HEAD if styleInHead === true', function() {
-      const options = {
-        cellHeight: 80,
-        verticalMargin: 10,
-        float: false,
-        styleInHead: true
-      };
-      GridStack.init(options);
-      expect(document.querySelector("style")).toBe(null);
-    });
-  });
+  }); 
 
   describe('grid.enableMove', function() {
     beforeEach(function() {
--- a/spec/gridstack-spec.ts
+++ b/spec/gridstack-spec.ts
@@ -1370,41 +1370,24 @@ describe('gridstack', function() {
     afterEach(function() {
       document.body.removeChild(document.getElementById('gs-cont'));
     });
-    it('should add STYLE to parent node as a default', function() {
-      var options = {
+    it('should not add STYLE to parent node', function() {
+      const options = {
         cellHeight: 80,
         verticalMargin: 10,
         float: false,
       };
-      var grid = GridStack.init(options);
-      expect((grid as any)._styles.ownerNode.parentNode.tagName).toBe('DIV'); // any to access private _styles
+      GridStack.init(options);
+      expect(document.querySelector("style")).toBe(null);
     });
-    it('should add STYLE to HEAD if styleInHead === true', function() {
-      var options = {
+    it('should not add STYLE to HEAD if styleInHead === true', function() {
+      const options = {
         cellHeight: 80,
         verticalMargin: 10,
         float: false,
         styleInHead: true
       };
-      var grid = GridStack.init(options);
-      expect((grid as any)._styles.ownerNode.parentNode.tagName).toBe('HEAD'); // any to access private _styles
-    });
-  });
-
-  describe('grid.opts.styleInHead', function() {
-    beforeEach(function() {
-      document.body.insertAdjacentHTML('afterbegin', gridstackHTML);
-    });
-    afterEach(function() {
-      document.body.removeChild(document.getElementById('gs-cont'));
-    });
-    it('should add STYLE to parent node as a default', function() {
-      var grid = GridStack.init();
-      expect((grid as any)._styles.ownerNode.parentNode.tagName).toBe('DIV');
-    });
-    it('should add STYLE to HEAD if styleInHead === true', function() {
-      var grid = GridStack.init({styleInHead: true});
-      expect((grid as any)._styles.ownerNode.parentNode.tagName).toBe('HEAD');
+      GridStack.init(options);
+      expect(document.querySelector("style")).toBe(null);
     });
   });
 
--- a/src/gridstack.scss
+++ b/src/gridstack.scss
@@ -52,6 +52,8 @@ $animation_speed: .3s !default;
   &.size-to-content:not(.size-to-content-max) > .grid-stack-item-content {
     overflow-y: hidden;
   }
+
+  height: var(--gs-cell-height);
 }
 
 .grid-stack-item {
@@ -98,6 +100,19 @@ $animation_speed: .3s !default;
   > .ui-resizable-sw { cursor: sw-resize; width: 20px; height: 20px;}
   > .ui-resizable-w  { cursor: w-resize;  width: 10px; top: 15px; bottom: 15px; }
 
+  > .ui-resizable-n {
+    top: var(--gs-item-margin-top);
+  }
+  > .ui-resizable-s, > .ui-resizable-se, > .ui-resizable-sw {
+    bottom: var(--gs-item-margin-bottom);
+  }
+  > .ui-resizable-ne, > .ui-resizable-e, > .ui-resizable-se {
+    right: var(--gs-item-margin-right);
+  }
+  > .ui-resizable-nw, > .ui-resizable-w, > .ui-resizable-sw {
+    left: var(--gs-item-margin-left);
+  }
+  
   &.ui-draggable-dragging {
     &> .ui-resizable-handle {
       display: none !important;
@@ -156,47 +171,12 @@ $animation_speed: .3s !default;
   width: 100%;
 }
 
-.grid-stack > .grid-stack-item {
-  height: var(--gs-cell-height);
+.grid-stack {
+  > .grid-stack-item > .grid-stack-item-content, > .grid-stack-placeholder > .placeholder-content {
+    top: var(--gs-item-margin-top);
+    right: var(--gs-item-margin-right);
+    bottom: var(--gs-item-margin-bottom);
+    left: var(--gs-item-margin-left);
+  }
 }
 
-.grid-stack > .grid-stack-item > .grid-stack-item-content {
-  top: var(--gs-item-margin-top);
-  right: var(--gs-item-margin-right);
-  bottom: var(--gs-item-margin-bottom);
-  left: var(--gs-item-margin-left);
-}
-
-.grid-stack > .grid-stack-placeholder > .placeholder-content {
-  top: var(--gs-item-margin-top);
-  right: var(--gs-item-margin-right);
-  bottom: var(--gs-item-margin-bottom);
-  left: var(--gs-item-margin-left);
-}
-
-.grid-stack > .grid-stack-item > .ui-resizable-n {
-  top: var(--gs-item-margin-top);
-}
-.grid-stack > .grid-stack-item > .ui-resizable-s {
-  bottom: var(--gs-item-margin-bottom);
-}
-.grid-stack > .grid-stack-item > .ui-resizable-ne {
-  right: var(--gs-item-margin-right);
-}
-.grid-stack > .grid-stack-item > .ui-resizable-e { 
-  right: var(--gs-item-margin-right);
-}
-.grid-stack > .grid-stack-item > .ui-resizable-se {
-  right: var(--gs-item-margin-right);
-  bottom: var(--gs-item-margin-bottom);
-}
-.grid-stack > .grid-stack-item > .ui-resizable-nw {
-  left: var(--gs-item-margin-left);
-}
-.grid-stack > .grid-stack-item > .ui-resizable-w {
-  left: var(--gs-item-margin-left);
-}
-.grid-stack > .grid-stack-item > .ui-resizable-sw {
-  left: var(--gs-item-margin-left);
-  bottom: var(--gs-item-margin-bottom);
-}
--- a/src/gridstack.scss
+++ b/src/gridstack.scss
@@ -155,3 +155,48 @@ $animation_speed: .3s !default;
 .gs-1 > .grid-stack-item {
   width: 100%;
 }
+
+.grid-stack > .grid-stack-item {
+  height: var(--gs-cell-height);
+}
+
+.grid-stack > .grid-stack-item > .grid-stack-item-content {
+  top: var(--gs-item-margin-top);
+  right: var(--gs-item-margin-right);
+  bottom: var(--gs-item-margin-bottom);
+  left: var(--gs-item-margin-left);
+}
+
+.grid-stack > .grid-stack-placeholder > .placeholder-content {
+  top: var(--gs-item-margin-top);
+  right: var(--gs-item-margin-right);
+  bottom: var(--gs-item-margin-bottom);
+  left: var(--gs-item-margin-left);
+}
+
+.grid-stack > .grid-stack-item > .ui-resizable-n {
+  top: var(--gs-item-margin-top);
+}
+.grid-stack > .grid-stack-item > .ui-resizable-s {
+  bottom: var(--gs-item-margin-bottom);
+}
+.grid-stack > .grid-stack-item > .ui-resizable-ne {
+  right: var(--gs-item-margin-right);
+}
+.grid-stack > .grid-stack-item > .ui-resizable-e { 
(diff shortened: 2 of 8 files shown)


src/gridstack.scss
```diff
@@ -155,3 +155,48 @@ $animation_speed: .3s !default;
...
+}
+
+.grid-stack > .grid-stack-item > .grid-stack-item-content {
+  top: var(--gs-item-margin-top);
+  right: var(--gs-item-margin-right);
+  bottom: var(--gs-item-margin-bottom);
+  left: var(--gs-item-margin-left);
+}
+
+.grid-stack > .grid-stack-placeholder > .placeholder-content {
```
adumesny, line 170: combine with 163. this is scss after all and can be moved as well

src/gridstack.scss
```diff
@@ -155,3 +155,48 @@ $animation_speed: .3s !default;
 .gs-1 > .grid-stack-item {
   width: 100%;
 }
+
+.grid-stack > .grid-stack-item {
```
adumesny, line 159: this can be more to line 42 content

src/gridstack.scss
```diff
@@ -155,3 +155,48 @@ $animation_speed: .3s !default;
...
+
+.grid-stack > .grid-stack-placeholder > .placeholder-content {
+  top: var(--gs-item-margin-top);
+  right: var(--gs-item-margin-right);
+  bottom: var(--gs-item-margin-bottom);
+  left: var(--gs-item-margin-left);
+}
+
+.grid-stack > .grid-stack-item > .ui-resizable-n {
+  top: var(--gs-item-margin-top);
```
adumesny, line 178: move to lines starting at 92

src/gridstack.ts
```diff
@@ -1547,78 +1538,31 @@ export class GridStack {
...
-    if (maxH > this._styles._max) {
-      const getHeight = (rows: number): string => (cellHeight * rows) + cellHeightUnit;
-      for (let i = this._styles._max + 1; i <= maxH; i++) { // start at 1
-        Utils.addCSSRule(this._styles, `${prefix}[gs-y="${i}"]`, `top: ${getHeight(i)}`);
-        Utils.addCSSRule(this._styles, `${prefix}[gs-h="${i + 1}"]`, `height: ${getHeight(i + 1)}`); // start at 2
-      }
-      this._styles._max = maxH;
-    }
+    // Set CSS var of cell height
+    this.setVar(this.el.parentElement, "--gs-cell-height", `${this.opts.cellHeight}${this.opts.cellHeightUnit}`);
```
adumesny, line 1559: can we use this.el instead ? the parent div could contain multiple grids which will break if they have different options. the old stylesheet had unique grid id (old 1578) for style to only apply to them...
this will also be needed for nested grid (parent of a grid is a gridItemContent)

src/gridstack.ts
```diff
@@ -1547,78 +1538,31 @@ export class GridStack {
...
 
-    const cellHeight = this.opts.cellHeight as number;
-    const cellHeightUnit = this.opts.cellHeightUnit;
-    const prefix = `.${this._styleSheetClass} > .${this.opts.itemClass}`;
-
-    // create one as needed
-    if (!this._styles) {
-      // insert style to parent (instead of 'head' by default) to support WebComponent
-      const styleLocation = this.opts.styleInHead ? undefined : this.el.parentNode as HTMLElement;
-      this._styles = Utils.createStylesheet(this._styleSheetClass, styleLocation, {
```
adumesny, line 1584: delete that util if not already done

spec/gridstack-spec.ts
```diff
@@ -1370,41 +1370,24 @@ describe('gridstack', function() {
     afterEach(function() {
       document.body.removeChild(document.getElementById('gs-cont'));
     });
-    it('should add STYLE to parent node as a default', function() {
-      var options = {
+    it('should not add STYLE to parent node', function() {
```
adumesny, line 1373: remove those test (no need to test removed code)

src/gridstack.ts
```diff
@@ -1547,78 +1538,31 @@ export class GridStack {
...
-        nonce: this.opts.nonce,
```
adumesny, line 1585: nonce: does that need to be done elsewhere now ?

src/gridstack.ts
```diff
@@ -1547,78 +1538,31 @@ export class GridStack {
...
-      });
-      if (!this._styles) return this;
-      this._styles._max = 0;
-
-      // these are done once only
-      Utils.addCSSRule(this._styles, prefix, `height: ${cellHeight}${cellHeightUnit}`);
```
adumesny, line 1591: another util to remove...

src/gridstack.ts
```diff
@@ -1677,17 +1621,27 @@ export class GridStack {
...
-    if (n.x !== undefined && n.x !== null) { el.setAttribute('gs-x', String(n.x)); }
-    if (n.y !== undefined && n.y !== null) { el.setAttribute('gs-y', String(n.y)); }
-    n.w > 1 ? el.setAttribute('gs-w', String(n.w)) : el.removeAttribute('gs-w');
-    n.h > 1 ? el.setAttribute('gs-h', String(n.h)) : el.removeAttribute('gs-h');
+  /**
+   * Call to write position x,y,w,h attributes back to element
+   * In addition, updates the inline top/height inline style as well
+   * @internal
+   */
+  protected _writePosAttr(el: HTMLElement, node: GridStackNode): GridStack {
```
adumesny, line 1629: didn't need to change n to node. now all looks different (and I use shortcut n(ode) or w(widget) all the time...). please revert.

src/gridstack.ts
```diff
@@ -1677,17 +1621,27 @@ export class GridStack {
...
+    if (node.x !== undefined && node.x !== null) { el.setAttribute('gs-x', String(node.x)); }
+    if (node.y !== undefined && node.y !== null) { el.setAttribute('gs-y', String(node.y)); }
+    node.w > 1 ? el.setAttribute('gs-w', String(node.w)) : el.removeAttribute('gs-w');
+    node.h > 1 ? el.setAttribute('gs-h', String(node.h)) : el.removeAttribute('gs-h');
+    // Avoid overwriting the inline style of the draggable element, but update the placeholder
+    if (!node._moving || this._placeholder === el) {
+      // Set inline style, refer CSS variables
```
adumesny, line 1636: ok, so instead of having global styles at 1617 (written once) you now have 2 inline styles (and 2 more for column if we did the same) PER instance and you think that is actually faster for large number of items ?
I think the opposite.

Also are the var used as is or actual values ? meaning if I change cellHeight so we need to update ALL items styles all over again ? just making sure it write el.style.top = calc(1 * var(--gs-cell-height))

Also technically we don't need gs-y | gs-h anymore since they don't refer to styles now so we could remove.
and you need to remove style when h=1 like I did for gs-h

src/gridstack.ts
```diff
@@ -2306,7 +2260,7 @@ export class GridStack {
           this.resizeToContentCheck(false, node);
           if (subGrid) {
             subGrid.parentGridNode = node;
-            if (!subGrid.opts.styleInHead) subGrid._updateStyles(true); // re-create sub-grid styles now that we've moved
+            subGrid._updateStyles(); // re-create sub-grid styles now that we've moved
```
adumesny, line 2263: might not be needed. all _updateStyles() need to be verified if needed to be called actually... sicne the external file isn't used now.

src/gridstack.ts
```diff
@@ -2493,7 +2447,7 @@ export class GridStack {
     node.el = this.placeholder;
     node._lastUiPosition = ui.position;
     node._prevYPix = ui.position.top;
-    node._moving = (event.type === 'dragstart'); // 'dropover' are not initially moving so they can go exactly where they enter (will push stuff out of the way)
+    node._moving = (event.type === 'dragstart' || event.type === 'resizestart'); // 'dropover' are not initially moving so they can go exactly where they enter (will push stuff out of the way)
```
adumesny, line 2450: that logic could break other code... might need to check _sizing or something in your code instead. please revert.

adumesny (commented):
I really like the idea of using variables to limit changes, but having 2 styles per items now (we can optimized when 1x1 items like I did) makes me wonder if this will indeed be faster for rendering.

I've wanted to do the same for columns as we only generate CSSS for 1 & 12 and many people forget to include 2-11 and wonder why things don't render. So I would like to do this for column more so actually.

this means complete removal of gs-x,y,w,h as those are only used to refer to classes anyway.

adiessl:
I am putting this here as you mentioned it in your comment above @adumesny: we are for example using the attributes `gs-x`, `gs-y`. `gs-h` and `gs-w` to style grid cells based on their dimensions using purely CSS. Should these no longer exist, it would be much more difficult to achieve the same dynamic styling as it would involve additional code to do so.

It should be no problem, however, if they were replaced with **unitless** CSS custom properties like `--gs-h` and so on. If they are not unitless, say `--gs-h` would be `100px` or `33%`, it would not work. So please keep in mind that there might be people using these attributes and removing them might cause breaking changes. Thank you.

adumesny:
@adiessl how would unitless CSS work for you ? wouldn't all have --gs-h then and would would you tell them appart (height of 1 vs 2) ?
so you are saying you style a cell h=1 and h=2 (or width) differently ? Maybe it could be accomplished using custom class per types instead, or key off class.height value it will become ?

adiessl:
I forked the official Fiddle and amended the CSS to show how we are using the attributes at the moment to display the grid cells' dimensions: https://jsfiddle.net/2y7w3jL0/

Unfortunately I was mistaken about being able to use unitless CSS custom properties. A CSS custom property cannot be used as `content` of an `after` element.

src/gridstack.ts
lmartorella: I've noticed that the "_moving" is already used for both move and resize gestures. However, let's keep them separated for this specific item about overriding the inline style on the fly.

src/gridstack.ts
lmartorella: The formula is the same, don't need to update all nodes when the cell-height changes. The CSS vars will be recalculated by the browser when the var changes.
Using inline style (even if this is a formula like that) is O(N) where N = number of widgets in the gridstack. Instead, using row/column rules is O(R+C) where R and C are the number of rows and columns, regardless the nodes actually on screen.
Using a finer resolution in the grid causes more rules to be put in CSS, even if only one node is on screen.

Given the intrinsic performance used by displaying nodes, having a O(N) rule in this case is more linear to the screen DOM complexity.

> Also technically we don't need gs-y | gs-h anymore since they don't refer to styles

I'm not sure if such attributes are used by the library or by other applications to do some logic, so probably it is better to leave them to avoid regressions.

adumesny: actually O(N) vs O(R!) - really 2xN vs 2xR! but yeah as I've seen people use finner resolutions, even down to pixels so this then makes a lot more sense.

adumesny:
@adiessl cool example (missing 1x1 case when no attr). we should add that as a demo.

src/gridstack.scss
```diff
@@ -98,6 +100,19 @@ $animation_speed: .3s !default;
   > .ui-resizable-sw { cursor: sw-resize; width: 20px; height: 20px;}
   > .ui-resizable-w  { cursor: w-resize;  width: 10px; top: 15px; bottom: 15px; }
 
+  > .ui-resizable-n {
+    top: var(--gs-item-margin-top);
```
adumesny, line 104: that conflicts with 95. need to figure out why I have 0 while you have '5' - either way we don't want 2 rules...
same with rest below...

src/gridstack.scss
```diff
@@ -98,6 +100,19 @@ $animation_speed: .3s !default;
...
+  }
+  > .ui-resizable-s, > .ui-resizable-se, > .ui-resizable-sw {
```
adumesny, line 106: 3 lines (3 rules) please. hard to read

src/gridstack.scss
```diff
@@ -155,3 +170,13 @@ $animation_speed: .3s !default;
 .gs-1 > .grid-stack-item {
   width: 100%;
 }
+
+.grid-stack {
+  > .grid-stack-item > .grid-stack-item-content, > .grid-stack-placeholder > .placeholder-content {
```
adumesny, line 175: 2 lines.

src/gridstack.ts
```diff
@@ -1317,7 +1308,7 @@ export class GridStack {
           // restore any sub-grid back
           if (n.subGrid?.el) {
             itemContent.appendChild(n.subGrid.el);
-            if (!n.subGrid.opts.styleInHead) n.subGrid._updateStyles(true); // force create
+            n.subGrid._updateStyles();
```
adumesny, line 1311: again if we skipped redoing styles that were global, then most likely we don't need to do anything... need to go though all calls and only call when what we upate changes... 
ONLY when cellHeigt OR margin is changed, OR number of rows for grid total height (might need to tweak that)

src/gridstack.ts
```diff
@@ -1677,17 +1621,27 @@ export class GridStack {
...
+  protected _writePosAttr(el: HTMLElement, n: GridStackNode): GridStack {
     if (n.x !== undefined && n.x !== null) { el.setAttribute('gs-x', String(n.x)); }
     if (n.y !== undefined && n.y !== null) { el.setAttribute('gs-y', String(n.y)); }
     n.w > 1 ? el.setAttribute('gs-w', String(n.w)) : el.removeAttribute('gs-w');
     n.h > 1 ? el.setAttribute('gs-h', String(n.h)) : el.removeAttribute('gs-h');
+    // Avoid overwriting the inline style of the element during drag/resize, but always update the placeholder
+    if ((!n._moving && !n._resizing) || this._placeholder === el) {
+      // Set inline style, refer CSS variables
+      el.style.top = `calc(${n.y} * var(--gs-cell-height))`;
+      el.style.height = `calc(${n.h} * var(--gs-cell-height))`;
```
adumesny, line 1638: remove height style when =1 (set globally once) - lot less rules when you have a ton of 1x1 elements

adumesny (commented):
do you want to do column as well to match ?

src/gridstack.ts
lmartorella: Well, my changes are quite conservative, I don't have the full knowledge about the side effects expected by this call (i.e. _updateContainerHeight)

src/gridstack.scss
lmartorella: This comes from the old code:

```ts
      Utils.addCSSRule(this._styles, `${prefix} > .ui-resizable-n`, `top: ${top};`);
```

src/gridstack.ts
lmartorella: Same as above, since my changes are quite conservative, I don't feel that the performances are hurting so much in removing the line. Probably better to split _updateStyles and extract _updateContainerHeight from it?

src/gridstack.ts
lmartorella: Having the rule always set, even in 1x1 case, is IMO more readable and more robust to future refactoring. However, pushed the required change.

src/gridstack.ts
adumesny: fair enough. but if there was a condition styleInHead then that means it's location based and therefore entire thing doesn't apply.

src/gridstack.scss
adumesny: yep ok, so replace my 0 with your var then (just one css rule). all your resize should merge into my existing rules.

src/gridstack.ts
```diff
@@ -1635,7 +1635,12 @@ export class GridStack {
     if ((!n._moving && !n._resizing) || this._placeholder === el) {
       // Set inline style, refer CSS variables
       el.style.top = `calc(${n.y} * var(--gs-cell-height))`;
-      el.style.height = `calc(${n.h} * var(--gs-cell-height))`;
+      if (n.h !== 1) {
```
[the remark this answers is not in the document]
lmartorella: This will trigger the

> Extract the assignment of "el.style.height" from this expression.sonarlint(typescript:S1121)

sonar lint issue, part of the recommended TS set.
I quite agree to that, being a "delete" part of else branch as well.
A better way is probably trying to assign an empty string to height, or undefined.

adumesny:
changes in #2874 need to be cherry picked too.

src/gridstack.ts
```diff
@@ -1547,78 +1538,31 @@ export class GridStack {
...
-  protected _removeStylesheet(): GridStack {
-
-    if (this._styles) {
-      const styleLocation = this.opts.styleInHead ? undefined : this.el.parentNode as HTMLElement;
-      Utils.removeStylesheet(this._styleSheetClass, styleLocation);
-      delete this._styles;
-    }
-    return this;
+  private setVar(el: HTMLElement, varName: string, varValue: string) {
+    el.style.setProperty(varName, varValue);
```
adumesny, line 1542: no need for a 1 liner utility that simple.

adumesny:
@lmartorella this review needs to be rebased, and I cherry picked to try and fix a few things, but on push it created `feat/static-css` instead I wanted to update your review and give you credit for the changes. Can you rebase so I can merge it in and ew can iterate on it further ?

dory2380:
Now that the sass is removed entirely, how can the community add custom styling for themeing without the sass files?

I get that it working right out of the the box is great, but the open source aspect of being able to customise with our own sass variables is lost now that only the generated css is provided :(*

PR #2905, opened 2024-12-29
New demo - Vue3 render functions with GridStack.renderCB

abulka:
### Description

New demo of Vue3 render functions using GridStack.renderCB

Benefits 
- less code than using GridStack.addRemoveCB, since GridStack creates the 2 needed parent divs + classes
- more information passed to GridStack.renderCB() - (entire GridStackWidget is passed so you can use id or some other field as logic) 
- less templating needed (no need for div class="grid-stack-item wrapper) in each Vue component 
- fixes {{ info }} display

### Checklist

Checklist N/A


adumesny:
thank you for making this demo! let's make it linked from the index page as well.

PR #2921, opened 2025-01-08
Replaced initMouseEvent with MouseEvent constructor and added composed: true

Taal91:
### Description
The deprecated method initMouseEvent has been replaced by the MouseEvent constructor. Additionally, compose: true has been added to allow gridstack.js to be used inside of a web component with touch events.

### Checklist


adumesny:
thanks for making this change

h4de5:
it seems this was not part of the latest release 11.3.
any chance you can estimate when this will be published?

adumesny:
It should be there. I merged it. Will have to check when I'm back on computer...

h4de5:
it seems files on npmjs.com differ from the one in the github release.

https://www.npmjs.com/package/gridstack?activeTab=code

h4de5:
hm. yes this is true. 
But I mean in the version on npmjs.com none have been replaced. both places in dd-touch still use the .createEvent method.

Issue #2671, opened 2024-04-30
Change event not firing on nested grids (resize, move...)


ddowie:
Hi,

does "change" event work on nested grid items?
In my situation change event is triggered only when moving or resizing parent (1st level) grid items.

## Steps to reproduce
```ts
var items = [{"x":0,"y":0,"w":6,"id":"5352223231011-3142222-410151-132321-200104202021152022100","subGridOpts":{"children":[]}},{"x":6,"y":0,"w":6,"id":"133014223145-112141-4112121-120214-43422114222221410","subGridOpts":{"children":[{"id":"01344140020-1013320-4151323-1521411-1140103111252101411","subGridOpts":{"children":[]},"x":0,"y":0},{"id":"412141422121-102202-42145-12152321-5223321101112214225","subGridOpts":{"children":[]},"x":3,"y":0}]}},{"x":0,"y":1,"w":12,"id":"1122021332120-2201014-4151315-1521145-12102130152142222321","subGridOpts":{"children":[]}}];

let options = { // main grid options
    resizable: {
        handles: 'se'
    },
    margin: 8,
    minRow: 2, // don't collapse when empty
    acceptWidgets: true,
    subGridDynamic: false,
    subGridOpts: subOptions // all sub grids will default to those
};


var grid = GridStack.init(options);
grid.load(items);

grid.on('change', function (event, el) {
      serializeFullGrid(grid);
  });

```

adumesny:
fixed in v12.1 - don't forget to donate if you fid this lib useful!

PR #2938, opened 2025-02-07
Simplify react lib design

Aysnine:
### Description

- Simple demo that use id selector
- Advanced demo with custom dynamic components

### Checklist


Aysnine:
```tsx
function App() {
  const [uncontrolledInitialOptions] = useState<GridStackOptions>({
    // ...
    children: [
      { id: "item1", h: 2, w: 2, x: 0, y: 0 },
      { id: "item2", h: 2, w: 2, x: 2, y: 0 },
    ],
  });

  return (
    <GridStackProvider initialOptions={uncontrolledInitialOptions}>
      <Toolbar />

      <GridStackRender>
        <GridStackItem id="item1">
          <div>hello</div>
        </GridStackItem>

        <GridStackItem id="item2">
          <div>grid</div>
        </GridStackItem>
      </GridStackRender>
    </GridStackProvider>
  );
}
```

react/lib/grid-stack-context.ts
```diff
@@ -1,35 +0,0 @@
-import type { GridStack, GridStackOptions, GridStackWidget } from "gridstack";
```
adumesny, line 1: hey don't delete those files. I actually wanted to start shipping them like I do for /angular 
I started looking at the code and was planning to do this weekend... I wanted to match better what I did for angular wrapper (extending WidgetItem to have similar to 'selector' + 'input' which are angular centric but conceptually the same as name + props you have.

react/src/lib/grid-stack-context.ts
```diff
@@ -0,0 +1,33 @@
+import type {
```
adumesny, line 1: oh I see you moved code here... quite different. will have to take a look so we can make this maybe official soon and publish as NPM.

adumesny:
if you look at the [angular version](https://github.com/gridstack/gridstack.js/tree/master/angular), this is what I have
```html
<gridstack [options]="gridOptions"></gridstack>
```
Is there a reason React can't match exactly ? the above is very simple DOM and provides the FULL power of GS as it doesn't try to create DOM children directly (and sync states between 2 frameworks) ?

in the DOM 'simple way (not recommended) that closer to your simple case above I also support
```html
<gridstack [options]="gridOptions" (changeCB)="onChange($event)">
   @for (n of items; track n.id) {
    <gridstack-item [options]="n">Item {{n.id}}</gridstack-item>
  }
</gridstack>
```

so I have <gridstack> and <gridstack-item> components.

react/src/lib/grid-stack-context.ts
Aysnine: Maybe we need to agree on the location of the libs, I didn't check how the wrappers for the other frameworks were packaged, so I temporarily moved the folders to make it easier for me to copy the complete code between multiple demo codebases

Aysnine:
I don't have experience with angular so I didn't check its example. But it seems necessary to learn angular examples, I'll go back and confirm.

react/src/lib/grid-stack-context.ts
adumesny: exact location doesn't matter. I just need to publish those as part of the NPM package so users can use the classes directly and get version fixes... angular happens to use angular\projects\lib\src\lib but what you have is fine.

we need to try and match angular wrapper syntax and simplicity whiel fully supporting nested and custom component inside grid items (gs cna handle the containers) and allow darg&drop between grid without re-creating the expensive content (you app components). that's what I did for Angular.

thank you for creating React version as I know it.

Aysnine:
I'm gonna set up this pr as a draft. We can align the lib design here.

react/src/lib/grid-stack-context.ts
Aysnine: Got it.

I haven't specifically focused on the darg&drop between grid issue, but I find it works well in this pr demo.

https://gridstack-react.pages.dev/

adumesny:
| Toolbar, which doesn't want to be in the gridstack container dom but needs to access the gridstack instance

why ? the gristack init params says if it accepts external widgets. then STATIC `GridStack.setupDragIn()` is used to init toolbar, but it doesn't need to access the instance (there could be multiple). see two.html

Aysnine:
> why ? the gristack init params says if it accepts external widgets (add enter/drop events). then STATIC GridStack.setupDragIn() is used to init toolbar (adds drag events), but it doesn't need to access the instance (there could be multiple). see two.html

My example here may not be appropriate. It's more like actions like save / restore require access to instance.

The toobar I have here is not quite like the one in two.html, it's more like it's made purely to give an example of how to access the gridstack through the react context. the toolbar in two.html I haven't tried yet, it's the one that's missing from this pr, I'll follow up on that.

> why having to maintain external state ? gridstack dom elements already have a gridstack pointer (and item GridItemHTMLElement have gridstackNode) and in the angular case I also store the Angular component pointers _gridComp | _gridItemComp. so from Class or DOM I can access either one. When DOM get destroyed, those are also gone - see ngOnDestroy

As I follow the component-first principle (instead of dom-first), the GridStackProvider doesn't just store the state, but also takes care of handling its lifecycle. And it provides a more standard react context way for other components to access the gridstack instance instead of getting it from the dom.

> in the angular case I use addRemoveCB (see gsCreateNgComponents) rather than the lower renderCB because angular component (your app children) can only live inside other angular components, so <gridstack> and <gridstack-item> are both angular components, but that's an angular restrictions and I'm guessing React can be hosted inside any plain DOM. So yeah I added renderCB exactly for that case...

You've reminded me that I haven't focused on ways other than renderCB, but renderCB is enough for me at the moment. I'll go back and check.

> you can at the very least use the same dom name and [options] so the code looks mostly the same.

LOL. That's what I'm struggling with, I'm finding places in the code where it's <gridstack-xxx>, which is a word, and places where it's type GridStack. In some places it's type GridStack, two words. Since it's common in react to use big humps for components, and since word spelling is easy to recognize, I followed the name type GridStack.

> We also need to talk about how custom comp are in the JSON. the old content string was legacy stuff. In angular I subclasses and added 2 params selector + input (very angular term, the second matching component @Input vs @Output fields for params). Thinking I may want to make content be a custom type (liek generic 'data' typically is) that default to string for legacy but can be our JSON app framework field of choices or something...

I think selector + input that's a good idea.

Also, in this pr, I removed the react implementation of JSON content from the lib to make it more pure. Putting the removed code implementation into advanced is an example of this.

If selector + input is implemented in gridstack, I just need to update my grid-stack-item component implementation for quick adaptation.

Aysnine:
> and THANK YOU for working with me on a native React wrapper as I always wanted to get this out but don't use/know React

gridstack provides very good support for custom panels for my in-house projects, and it's worth providing a better-working react wrapper for this purpose :)

adumesny:
| It's more like actions like save / restore require access to instance

`GridStack.init()` or `GridStack.initAll()` return instances, and angualr wrapper `GridstackComponent` stores it when doing `this._grid = GridStack.init()` so no need to keep a global list (one fewer thing to get out of sync). it's up to the app if they want to track `<gridstack>` which has an angular way of doing.

Aysnine:
> If selector + input is implemented in gridstack

My expression was inaccurate, I meant to say that gridstack defines the option data type well.

> maybe make existing 'content'

What about considering new fields? Like `data` or `meta`? That way you don't have to break the `content`.

Aysnine:
@adumesny  I just realized a serious bug where <GridStackProvider /> assigns a value to GridStack.renderCB, and when there are multiple <GridStackProvider />'s, this causes GridStack.renderCB to be duplicated, resulting in a duplication of the id maintained in each Provider. element map maintained in each Provider is inaccurate.

Also, the id doesn't seem to be required for each item, which I hadn't considered.

adumesny:
internal ids have to be globaly unique (why it's a running static counter) as you can move a widget from one grid to another... user ids should be unique at least per grid so save/load can map to correct instance of a widget to update.

Aysnine:
> internal ids have to be globaly unique (why it's a running static counter) as you can move a widget from one grid to another... user ids should be unique at least per grid so save/load can map to correct instance of a widget to update.

Resolved.

Aysnine:
Another confusion:

```ts
const instance = GridStack.init(optionsRef.current, containerRef.current);
instance.on("dropped", (e) => {
  console.log("dropped", e);
  // ! subGrid not working
});
```

All event not receive about subGrid.

I'm stuck here now because I want to listen for grid changes and dynamically render the react component in the react advanced example, which is a more common scenario.

adumesny:
> All event not receive about subGrid.|

correct, subgrids do not currently propagate events to the top one (where you add the event handler). see #2671, which I could look into

> I'm stuck here now because I want to listen for grid changes and dynamically render the react component in the react advanced example, which is a more common scenario.

when you move an item from one grid to another (child or not) you should NOT have to re-render the component (as creating them can be expensive, making API calls all over, etc... in the anular wrapper and let GS re-parent things and everything is OK BECAUSE I don't have Angualr create the dom fomr state, but let GS drive the state and callback to wrapper to create component when needed.

Are you not doing that ? that's a must change IMO.

Aysnine:
| when you move an item from one grid to another (child or not) you should NOT have to re-render the component (as creating them can be expensive, making API calls all over, etc... in the anular wrapper and let GS re-parent things and everything is OK BECAUSE I don't have Angualr create the dom fomr state, but let GS drive the state and callback to wrapper to create component when needed.

That's actually what I was thinking, but it's different with dragIn (clone), which always requires the creation of a new component.

And dragIn to subGrid can't be listened to. So I implemented the rendering of the dragIn copied component in a different way, instead of relying on listening. But the downside is that it's not good for save/load.

adumesny:
| it's different with dragIn (clone), which always requires the creation of a new component.

right, drag froms idebar is always  apreview and actual comp needs to be created. instead it's cleaner, and what I have in angualr wrapper, to use the WidgetItem as what to create and use `GridStack.renderCB` or `GridStack.addRemoveCB` which is also used during reading, to create the right component bu storing custom info on what to create (angular 'selector' field). React should do the same.

I really want React and Angular wrapper doc to look and act the same. so again I suggest you study what I do and do the equicalent react code version if not totally weird for that framework. GS needs to maintain state and re-parent and resize things as it needs to, independ on the framework. the framework only job is to create the right element (or `<gridstack>` <gridstack-item>` parent as angualr needs entire hierachy to be ng components) instead of plain divs with classes.

Aysnine:
> I really want React and Angular wrapper doc to look and act the same.

Yes. I want to run /angular folder. But error when install deps:

```
➜ gridstack.js/angular master ✓ yarn
yarn install v1.22.22
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error Error: https://usw1.packages.broadcom.com/artifactory/api/npm/tis-npm-virtual/@angular/core/-/core-14.3.0.tgz: Request failed "403 Forbidden"
    at ResponseError.ExtendableBuiltin (/Users/jace/.cache/node/corepack/v1/yarn/1.22.22/lib/cli.js:696:66)
    at new ResponseError (/Users/jace/.cache/node/corepack/v1/yarn/1.22.22/lib/cli.js:802:124)
    at Request.<anonymous> (/Users/jace/.cache/node/corepack/v1/yarn/1.22.22/lib/cli.js:66750:16)
    at Request.emit (node:events:524:28)
    at module.exports.Request.onRequestResponse (/Users/jace/.cache/node/corepack/v1/yarn/1.22.22/lib/cli.js:142287:10)
    at ClientRequest.emit (node:events:524:28)
    at HTTPParser.parserOnIncomingClient (node:_http_client:716:27)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:117:17)
    at TLSSocket.socketOnData (node:_http_client:558:22)
    at TLSSocket.emit (node:events:524:28)
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Is this a package from a private registry?

adumesny:
| error Error: https://usw1.packages.broadcom.com/artifactory/api/npm/tis-npm-virtual/@angular/core/-/core-14.3.0.tgz

oops, sorry. let me fix that asap. yes, private repo...

adumesny:
sorry about private registry! fixed in #2977

Aysnine:
@adumesny  I have two questions about angular wrapper:

1. Is comp1 re-render when move item from grid a to grid b ?
2. What will be happen when move item from grid a to grid b and destroy grid a?

image

I'm wondering if the comp1 lifecycle goes with the grid, or with the dom ref, or something else.

adumesny:
1. no delete/re-create. you can see that using the /angular demo (yarn + yarn start), pick two grid demo, and look at the log while dragging an object B to the other grid
image

Aysnine:
> angular can have dynamic list of children and I'm pretty sure when GS manually reparents a child the parent 'container' doesn't try to destroy an old child...

I think this may be the essence of why react wrappers are not easy to design. react's rendering style dictates that item content cannot be designed as a react component that can be re-parented.

Same two questions but in react:

1. no delete/re-create.
2. item content (not item container by GS) will be destroy

That's why, I've seen previous GS react wrappers or examples attempting controlled components, not uncontrolled ones like mine. It's because they want to dispatch the react component with the help of the item meta in the options state to make sure the component looks re-parent.

I need to calm down a bit. 

While it occurred to me that I could use react's multiple createRoots to get each item content component off of its parent, this is not very performant and goes against the principle of a single root for react.

adumesny:
ok, so how do you want to proceed ? in the end when a widget is moved from a grid to another it is best to not recreate the expensive content (could be long API calls) and let GS just do a DOM re-parenting. this doesn't break Angular wrapper as GridItem comp (which then contains actual user comp) are created dynamically and will update parents list of children for destroy.

if that's not possible in React then so be it.

chungwong:
I gave the new design a spin but I didn't get far. Is it ready for testing?
The first error I am getting is [getContainerByWidgetId is unimplemented yet](https://github.com/Aysnine/gridstack-react/blob/main/src/lib/grid-stack-render-context.ts#L10)

AO19:
@Aysnine what is your status on this? Did you get more progress?

lelik-anabolek:
@Aysnine how is it going? Mb, do you need help?

adumesny:
according to Cursor AI, the current implementation is better than this changelist  (I think ability to re-parent like I do in Angular is too important) even though the DOM format is closer to HTML & Angular we have now...

Would like to get React experts advice on this:

[cursor_understanding_react_basics.md](https://github.com/user-attachments/files/21827263/cursor_understanding_react_basics.md)

### **Current Implementation is Better Because:**

1. **Handles React's Re-parenting Limitation**
   - The current Portal-based approach solves React's fundamental issue with DOM re-parenting
   - As discussed in the PR, React components get destroyed/recreated when moved between grids
   - Current implementation avoids this by using portals to render into GridStack-managed containers

2. **Production-Ready Features**
   - Supports all GridStack features (nested grids, responsive layouts, drag-between-grids)
   - Proper memory management with WeakMaps
   - Error handling and edge cases covered

3. **Framework Alignment**
   - The maintainer (adumesny) wants to match the Angular wrapper pattern
   - Current approach provides similar functionality to the Angular implementation
   - Ready for NPM publishing

4. **Performance Considerations**
   - Avoids React component recreation during drag operations
   - Maintains component state during grid operations
   - Efficient container mapping system

demarchenac:
@adumesny Hello!
Fellow developer over here, I wanted to know if this is related to the work mentioned here [gridstack.js/react/README.md](https://raw.githubusercontent.com/gridstack/gridstack.js/refs/heads/master/react/README.md). Bc I somewhat confused into which is the recommended way to integrate grid-stack with react so I can render custom components as panels

adumesny:
@demarchenac /react (with readme) is the recommended way as this review has issues... I'm not a React person (I use Angular at work) and hoping to get community feedback but my points above are still valid.

adumesny:
closing as Claude is doing a better analysis and we now have uodated /react code to use.

PR #2949, opened 2025-02-17
fix(resize): update the grid to resize when initialised in the smaller screen

GowthamTG:
### Description
This change would help recalculate the grid-layout when it is getting initialised in the smaller screen

https://github.com/gridstack/gridstack.js/issues/2947

https://github.com/user-attachments/assets/7039b300-d881-4b76-b3c6-f3fa9ed45406

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -346,8 +346,6 @@ export class GridStack {
     opts = Utils.defaults(opts, defaults);
     this._initMargin(); // part of settings defaults...
 
-    // Now check if we're loading into 1 column mode FIRST so we don't do un-necessary work (like cellHeight = width / 12 then go 1 column)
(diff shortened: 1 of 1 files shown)


src/gridstack.ts
```diff
@@ -346,8 +346,6 @@ export class GridStack {
     opts = Utils.defaults(opts, defaults);
     this._initMargin(); // part of settings defaults...
 
-    // Now check if we're loading into 1 column mode FIRST so we don't do un-necessary work (like cellHeight = width / 12 then go 1 column)
-    this.checkDynamicColumn();
```
adumesny, line 350: this removes optimizations so this isn't the right fix even if it fixes the issue...

adumesny:
fixed the right way in v12

PR #2988, opened 2025-03-12
#2667 Fixed sidebar items not honouring gs-w

James-Linskell:
### Description
#2667 
These changes fix sidebar items not honoring gs-w and gs-h when dragging between multiple grids. When dragging between multiple grids, the size of the widget is limited by nodeBoundFix() based on number of columns. Eg. if you enter a grid which limits the gs-w attribute then enter another grid, I would expect the gs-w to honor the original size defined in the sidebar item, but instead it keeps the value which was limited in the previous grid.

These changes add another property for tracking the sidebar item's original position, so any number of grids can be entered and it still respects the original size, only limiting gs-w when appropriate. The existing _gridstackNodeOrig logic was not sufficient to handle the different behavior of the sidebar item.

These changes also fix another bug I discovered, which can be seen in the original example sandbox provided in #2667 . writePosAttr was writing the updated values to the _original sidebar element_, permanently altering gs-w, so any subsequent drags of the item would use the updated value. Sidebar items should not be mutable in this way.

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -2215,6 +2215,12 @@ export class GridStack {
           return false; // prevent parent from receiving msg (which may be a grid as well)
         }
 
+        // If sidebar item, restore the sidebar node size to ensure consistent behavior when dragging between grids
+        if (node?._sidebarOrig) {
+          node.w = node._sidebarOrig.w;
+          node.h = node._sidebarOrig.h;
+        }
+
         // fix #1578 when dragging fast, we may not get a leave on the previous grid so force one now
         if (node?.grid && node.grid !== this && !node._temporaryRemoved) {
           // console.log('dropover without leave'); // TEST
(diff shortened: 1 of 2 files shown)


src/gridstack.ts
```diff
@@ -2661,7 +2669,10 @@ export class GridStack {
       this._updateContainerHeight();
 
       const target = event.target as GridItemHTMLElement;// @ts-ignore
-      this._writePosAttr(target, node);
+      // Do not write sidebar item attributes back to the original sidebar el
+      if (!node._sidebarOrig) {
+        this._writePosAttr(target, node);
```
adumesny, line 2674: good fix to not write sidebar back, but still seems ,like we need to write to the dropped item... will have to debug.

src/types.ts
```diff
@@ -456,6 +456,8 @@ export interface GridStackNode extends GridStackWidget {
   _temporaryRemoved?: boolean;
   /** @internal true if we should remove DOM element on _notify() rather than clearing _id (old way) */
   _removeDOM?: boolean;
+  /** @internal original position/size of item if dragged from sidebar */
+  _sidebarOrig?: GridStackPosition;
```
adumesny, line 460: reluntant to add yet another data storage, but maybe this makes it clearer...

adumesny (approved):
thanks for the fix. dragging between 2 different restricted column count wasn't handled so this is good.

PR #3024, opened 2025-04-21
Support Grid Item Movement via External Element (outside Grid) Drag

imbansalaniket:
### Description

This pull request adds support for moving grid items when drag events are initiated from elements outside the grid. This enhancement allows developers to implement more interactive interfaces, such as custom drag handles or external controls that reposition grid items without direct interaction with the grid itself.

A specific scenario motivating this enhancement involves grid item titles that are displayed as popups when hovering over a grid item. These popup elements are appended to the HTML body (or a similar general DOM container) as per standard best practices for creating tooltips or popups, ensuring proper rendering and positioning. Since these popups exist outside the grid item’s DOM hierarchy, they cannot inherently trigger grid item movement.

To address this limitation, the grid’s drag-and-drop functionality has been extended to support external elements. This feature allows developers to dynamically reference and use such external elements (e.g., popup titles) as drag handles for moving grid items.

This solution leverages existing drag-and-drop infrastructure, ensuring seamless integration without introducing additional complexity.

A new demo, "Title Drag from Outside GridStack Item", has been added to illustrate this use case and demonstrate the versatility of this enhancement.

ExternalTitleDrag

### Checklist

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -148,6 +148,7 @@ v10.x supports a much richer responsive behavior, you can have breakpoints of wi
 - `scroll`?: boolean - default to 'true', enable or disable the scroll when an element is dragged on bottom or top of the grid.
 - `cancel`?: string - prevents dragging from starting on specified elements, listed as comma separated selectors (eg: '.no-drag'). default built in is 'input,textarea,button,select,option'
 - `helper`?: 'clone' | ((el: HTMLElement) => HTMLElement) - helper function when dragging side panel items that need to be cloned before dropping (ex: 'clone' or your own method)
+- `dragElements`?: HTMLElement[] - references to the elements (outside the grid item) to be used for dragging grid item
 
 ## Grid attributes
 
--- a/demo/demo.css
+++ b/demo/demo.css
@@ -59,6 +59,16 @@ h1 {
   cursor: move;
   min-height: 25px;
   background-color: #16af91;
+  &.outside {
+    display: none;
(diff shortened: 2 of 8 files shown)


doc/README.md
```diff
@@ -148,6 +148,7 @@ v10.x supports a much richer responsive behavior, you can have breakpoints of wi
 - `scroll`?: boolean - default to 'true', enable or disable the scroll when an element is dragged on bottom or top of the grid.
 - `cancel`?: string - prevents dragging from starting on specified elements, listed as comma separated selectors (eg: '.no-drag'). default built in is 'input,textarea,button,select,option'
 - `helper`?: 'clone' | ((el: HTMLElement) => HTMLElement) - helper function when dragging side panel items that need to be cloned before dropping (ex: 'clone' or your own method)
+- `dragElements`?: HTMLElement[] - references to the elements (outside the grid item) to be used for dragging grid item
```
adumesny, line 151: honestly this wouldn't work in general since you assume a fix set of elements (really a single one shared in your case) which doesn't work for dynamic elements. what happens when you add/remove/reparent items ?

adumesny:
not sure I understand why you need to be 'outside' the grid. you can use CSS to position you drag handle (and show/hide on mouse over like we do for resize handles) and still be per item and inside each grid - beside your showing the issue of the tab not moving along either (and disapearing) which isn't great.

Thanks for the effort but I cannot take this - see my other comments as well.

PR #3054, opened 2025-05-12
Avoid reflows via explicitly setting minRow

lmartorella:
### Description

If gridstack is used in a complex layout, the browser layout can be interrupted by a [reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) when the `getComputedStyle` function is called to ask for parent `min-height`.

image

To mitigate this in the `_updateContainerHeight`, it could be possible to give a priority if both `minRow` option and the `min-height` CSS property of the parent div are set. In this way, applications that sets `minRow` programmatically would never trigger the reflow.

Since `0` is the default value, and `null`/`undefined` values can be tricky to manage due to falsy/truly check in JS, to I would say that checking a non-zero/non-null value would be the solution. Users that wants "0" to be a valid row count and still disable the CSS check, could be pass something like `-1` to obtain the same result.

Thanks, L

### Checklist
- [N/A] Created tests which fail without the change (can't find a test for the `min-height` CSS prop, please let me know if you would like to have this tested)

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -116,7 +116,7 @@ gridstack.js API
 - `marginBottom`: numberOrString
 - `marginLeft`: numberOrString
 - `maxRow` - maximum rows amount. Default is `0` which means no max.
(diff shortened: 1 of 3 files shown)


adumesny:
you doc had typos but I will fix them... thanks.

PR #3070, opened 2025-06-09
fix: incorrect property name 'sizeToContent' when cleaning up invalid attributes

capdiem:
Setting `gs-size-to-content` on an element does not have the intended effect.


adumesny:
thank you!

capdiem:
@adumesny Thanks for merging the PR! Do you have an estimate on when the next release will be that includes this fix?

adumesny:
I try to have a bunch of fixes before I release (unless blocking) as too much overhead. likely a few weeks I'm afraid... you can patch the method locally in the meantime, while still using the offcial release.

Issue #3071, opened 2025-06-11
Layout breaks when grid is initilized with one column


FilipKujavec:
## Subject of the issue
Layout breaks when grid is initialized in one column through breakpoints. Both the order is wrong and when resizing, it doesnt properly restore the 12 column layout

## Your environment
* version of gridstack.js - 12.2.1
* which browser/OS: Firefox/Windows

## Steps to reproduce
https://jsfiddle.net/f3159znr/3/
Size output window down so its smaller than the breakpoint, then reload and resize it so its larger than the breakpoint

https://github.com/user-attachments/assets/107c1963-58f7-48d7-b06e-06c69a4c3337

## Expected behavior
Grid should behave like a grid initialized with 12 columns:

https://github.com/user-attachments/assets/b191cb24-771e-4e57-8425-a509cd2d73b3

I created a PR with a possible fix here: (PR following shortly)

adumesny:
your example has many issues, one of which is wrong widget DOM structure, so your widget touch and sometime render wrong... but with correct struture GridSTack:init() will initialise them all correctly, but you are missing DOM size info (you can call load() with all the sizes which calls update() on each.

you are mixing DOM load and json intialize... don't klnow where to start. That say you might have uncovered a valid bug that need fixing. will have to debug it later

![Image](https://github.com/user-attachments/assets/cc468142-b9a9-49a7-82f3-ec794620844e)

FilipKujavec:
https://jsfiddle.net/kp347naj/
Would this be a better example? This issue doesnt occur when using `load()` as it caches the layout in `columnChanged()`. But when using `addWidget()` or `makeWidget()` as they rely on `cacheOneLayout()` in `nodeBoundFix()`

PR #3072, opened 2025-06-11
fix: makeWidget breakpoint layout issue #3071

FilipKujavec:
### Description
See #3071 for description of issue. I added setting _loading to true in _prepareElement to fix the order issue and changed how we check if `node._id` is present (previously it would be false when called with the first element as it has an id of 0, not sure if this was wanted)

### Checklist
- [- ] Created tests which fail without the change (if possible)
- [X ] All tests passing (`yarn test`)
- [- ] Extended the README / documentation, if necessary

### DIFF
--- a/src/gridstack-engine.ts
+++ b/src/gridstack-engine.ts
@@ -417,14 +417,14 @@ export class GridStackEngine {
     // remember it's position & width so we can restore back (1 -> 12 column) #1655 #1985
     // IFF we're not in the middle of column resizing!
     const saveOrig = (node.x || 0) + (node.w || 1) > this.column;
-    if (saveOrig && this.column < this.defaultColumn && !this._inColumnResize && !this.skipCacheUpdate && node._id && this.findCacheLayout(node, this.defaultColumn) === -1) {
+    if (saveOrig && this.column < this.defaultColumn && !this._inColumnResize && !this.skipCacheUpdate && !isNaN(node._id) && this.findCacheLayout(node, this.defaultColumn) === -1) {
       const copy = {...node}; // need _id + positions
       if (copy.autoPosition || copy.x === undefined) { delete copy.x; delete copy.y; }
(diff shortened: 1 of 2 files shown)


src/gridstack-engine.ts
```diff
@@ -417,14 +417,14 @@ export class GridStackEngine {
     // remember it's position & width so we can restore back (1 -> 12 column) #1655 #1985
     // IFF we're not in the middle of column resizing!
     const saveOrig = (node.x || 0) + (node.w || 1) > this.column;
-    if (saveOrig && this.column < this.defaultColumn && !this._inColumnResize && !this.skipCacheUpdate && node._id && this.findCacheLayout(node, this.defaultColumn) === -1) {
+    if (saveOrig && this.column < this.defaultColumn && !this._inColumnResize && !this.skipCacheUpdate && !isNaN(node._id) && this.findCacheLayout(node, this.defaultColumn) === -1) {
```
adumesny, line 420: good catch!
node._id!==undefined or node._id?? might be less specific

FilipKujavec: Screenshot 2025-06-12 072132
`node._id` is currently defined as a number, but I can change it to `node._id??` to be more generic if wanted

src/gridstack.ts
```diff
@@ -1633,7 +1633,9 @@ export class GridStack {
     el.gridstackNode = node;
     node.el = el;
     node.grid = this;
+    this.engine._loading = true; // help with collision
```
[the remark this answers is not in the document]
FilipKujavec: removed it for now

src/gridstack-engine.ts
adumesny: yes please.

FilipKujavec: changed it

PR #3077, opened 2025-06-23
fix: incorrect node._id check

FilipKujavec:
### Description
Fixed a mistake I made in this PR https://github.com/gridstack/gridstack.js/pull/3072
`(node._id ?? false)` still is falsy when  `node_id = 0` 

### Checklist
- [-] Created tests which fail without the change (if possible)
- [-] Extended the README / documentation, if necessary

Also @adumesny  should I create a new issue for the incorrect order mentioned in #3071?

### DIFF
--- a/src/gridstack-engine.ts
+++ b/src/gridstack-engine.ts
@@ -417,7 +417,7 @@ export class GridStackEngine {
     // remember it's position & width so we can restore back (1 -> 12 column) #1655 #1985
     // IFF we're not in the middle of column resizing!
     const saveOrig = (node.x || 0) + (node.w || 1) > this.column;
-    if (saveOrig && this.column < this.defaultColumn && !this._inColumnResize && !this.skipCacheUpdate && (node._id ?? false) && this.findCacheLayout(node, this.defaultColumn) === -1) {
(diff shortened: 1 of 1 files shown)


src/gridstack-engine.ts
```diff
@@ -417,7 +417,7 @@ export class GridStackEngine {
     // remember it's position & width so we can restore back (1 -> 12 column) #1655 #1985
     // IFF we're not in the middle of column resizing!
     const saveOrig = (node.x || 0) + (node.w || 1) > this.column;
-    if (saveOrig && this.column < this.defaultColumn && !this._inColumnResize && !this.skipCacheUpdate && (node._id ?? false) && this.findCacheLayout(node, this.defaultColumn) === -1) {
+    if (saveOrig && this.column < this.defaultColumn && !this._inColumnResize && !this.skipCacheUpdate && node._id != null  && this.findCacheLayout(node, this.defaultColumn) === -1) {
```
adumesny, line 420: _id is internal to me, so it should always be set, or undefined. weird that you are checking (and should be !== null).
I would have to check why it isn't set (and why null works in your case ?

FilipKujavec: It is set, its just that a `_id` of 0 leads to it being falsy again as my earlier code was wrong. `!= null` just checks that its not null or undefined. If its never going to be null `!== undefined` would work fine as well
image

PR #3083, opened 2025-06-30
The raiser may get stuck on some element. #3082

### LINKED ISSUE #3082: The raiser may get stuck on some element.
## Subject of the issue
The raiser may get stuck on some item. To turn it on again, the user have to find this item and move the cursor out from this element. After that, the resizer will work properly.
Possible cause https://stackoverflow.com/questions/7448468/why-cant-i-reliably-capture-a-mouseout-event

## Your environment
 "gridstack": "^12.2.1",
Tested on
- Safari 15.6.1 the most easily reproducible 
- Chrome 137.0.7151.122 less frequently reproduced

## Steps to reproduce
Create long grid with subgrids.
Move cursor over resizable item
Swipe (scroll) fast to the top o bottom of the grid.
Perhaps the resizer is stuck on the first element. It will be impossible to resize of other elements.

## Expected behavior
 It is possible to resize of other elements.

catakot:
### Description
fix #3082

The _mouseOut may not clear the overResizeElement properly because there is already another component in the overResizeElement.

### DIFF
--- a/src/dd-resizable.ts
+++ b/src/dd-resizable.ts
@@ -132,7 +132,7 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   protected _mouseOver(e: Event): void {
     // console.log(`${count++} pre-enter ${(this.el as GridItemHTMLElement).gridstackNode._id}`)
     // already over a child, ignore. Ideally we just call e.stopPropagation() but see https://github.com/gridstack/gridstack.js/issues/2018
-    if (DDManager.overResizeElement || DDManager.dragElement) return;
+    if (DDManager.overResizeElement === this || DDManager.dragElement) return;
     DDManager.overResizeElement = this;
     // console.log(`${count++} enter ${(this.el as GridItemHTMLElement).gridstackNode._id}`)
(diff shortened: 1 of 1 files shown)


src/dd-resizable.ts
```diff
@@ -132,7 +132,7 @@ export class DDResizable extends DDBaseImplement implements HTMLElementExtendOpt
   protected _mouseOver(e: Event): void {
     // console.log(`${count++} pre-enter ${(this.el as GridItemHTMLElement).gridstackNode._id}`)
     // already over a child, ignore. Ideally we just call e.stopPropagation() but see https://github.com/gridstack/gridstack.js/issues/2018
-    if (DDManager.overResizeElement || DDManager.dragElement) return;
+    if (DDManager.overResizeElement === this || DDManager.dragElement) return;
     DDManager.overResizeElement = this;
```
adumesny, line 136: by doing this you are not clearing the other (if any) element style at line 148 this.el.classList.add('ui-resizable-autohide') so now 2 items are showing resize ?
would be relly helpfull to have a running example to test the issue first...

adumesny: will fix myself...

Issue #2937, opened 2025-02-06
React : multiple grids crashes


PlugNPush:
Hi,

I wanted to test gridstack for my project and made a very basic grid like this: 

```javascript
export function GridStackPage() {
  const [grid1Options] = useState(() => createGridOptions());
  const [grid2Options] = useState(() => createGridOptions());

  return (
    <div style={{ display: "flex", gap: "20px" }}>
      <div style={{ flex: 1 }}>
        <h3>Grid 1</h3>
        <GridStackProvider initialOptions={grid1Options}>
          <Toolbar gridNumber={1} />
          <GridStackRenderProvider>
            <GridStackRender componentMap={COMPONENT_MAP} />
          </GridStackRenderProvider>
        </GridStackProvider>
      </div>
      <div style={{ flex: 1 }}>
        <h3>Grid 2</h3>
        <GridStackProvider initialOptions={grid2Options}>
          <Toolbar gridNumber={2} />
          <GridStackRenderProvider>
            <GridStackRender componentMap={COMPONENT_MAP} />
          </GridStackRenderProvider>
        </GridStackProvider>
      </div>
    </div>
  );
}
```

Whenever I want to add a widget in them, it works fine for the last one but crashes for all the ones before (here grid 1).

The reason for the crash is : 

```
Uncaught Error: Widget container not found for id: widget-39vq52659p5
    children grid-stack-render.tsx:55
    GridStackRender grid-stack-render.tsx:47
    React 18
    current grid-stack-render-provider.tsx:35
    renderCB grid-stack-render-provider.tsx:41
    createWidgetDivs utils.ts:128
    addWidget gridstack.ts:470
    addSubGrid grid-stack-provider.tsx:60
    onClick GridStackDemo.tsx:210
```

It comes from GridStackRenderProvider, in 

```javascript
<GridStackRenderContext.Provider
        value={useMemo(
          () => ({
            getWidgetContainer: (widgetId: string) => {
              return widgetContainersRef.current.get(widgetId) || null;
            },
          }),
          // ! gridStack is required to reinitialize the grid when the options change
          // eslint-disable-next-line react-hooks/exhaustive-deps
          [gridStack]
        )}
      >
        <div ref={containerRef}>{gridStack ? children : null}</div>
      </GridStackRenderContext.Provider>
```

as widgetContainersRef.current.get(widgetId)  unexpectedly returns undefined.

I would appreciate some help here.

adumesny:
[...]

PR #3102, opened 2025-07-28
feat(test): add vitest setup and widget container tests

tearf001:
## Description
This PR fixes the issue where multiple GridStack instances would crash when using the React integration due to shared state in the static GridStack.renderCB callback (#2937).

Problem: The static GridStack.renderCB was being shared across all GridStack instances, causing widget container references to be mixed up between different grids. This led to crashes and incorrect behavior when multiple grids were present on the same page.

## Solution: Implemented a WeakMap-based approach to store widget containers separately for each GridStack instance:
•  Added a gridWidgetContainersMap WeakMap that uses the GridStack instance as the key
•  Each grid instance now maintains its own isolated widget container map
•  The WeakMap ensures automatic memory cleanup when grid instances are destroyed

## the fix snippets:
```ts
// Before: All grids shared the same widget containers
GridStack.renderCB = (el, widget) => {
  widgetContainersRef.current.set(widget.id, el); // Shared state
}

// After: Each grid has its own container map
GridStack.renderCB = (el, widget) => {
  if (widget.grid) {
    let containers = gridWidgetContainersMap.get(widget.grid);
    if (!containers) {
      containers = new Map();
      gridWidgetContainersMap.set(widget.grid, containers);
    }
    containers.set(widget.id, el); // Instance-specific state
  }
}
```
This ensures that multiple GridStack instances can coexist safely without interfering with each other's widget management.

## Checklist
- [x] All tests passing (yarn test)

## Additional notes:
•  Added comprehensive unit tests using Vitest to verify the WeakMap functionality
•  Tests cover: multi-instance isolation, proper cleanup, and widget container management
•  The solution maintains backward compatibility while fixing the multi-instance issue

### DIFF
--- a/react/yarn.lock
+++ b/react/yarn.lock
@@ -2,120 +2,289 @@
 # yarn lockfile v1
 
 
+"@asamuzakjp/css-color@^3.2.0":
+  version "3.2.0"
+  resolved "https://registry.npmmirror.com/@asamuzakjp/css-color/-/css-color-3.2.0.tgz#cc42f5b85c593f79f1fa4f25d2b9b321e61d1794"
+  integrity sha512-K1A6z8tS3XsmCMM86xoWdn7Fkdn9m6RSVtocUrJYIwZnFVkng/PvkEoWtOWmP+Scc6saYWHWZYbndEEXxl24jw==
+  dependencies:
+    "@csstools/css-calc" "^2.1.3"
+    "@csstools/css-color-parser" "^3.0.9"
+    "@csstools/css-parser-algorithms" "^3.0.4"
+    "@csstools/css-tokenizer" "^3.0.3"
+    lru-cache "^10.4.3"
+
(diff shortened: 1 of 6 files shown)


react/yarn.lock
```diff
@@ -2,121 +2,290 @@
 # yarn lockfile v1
 
 
+"@asamuzakjp/css-color@^3.2.0":
+  version "3.2.0"
+  resolved "https://registry.npmmirror.com/@asamuzakjp/css-color/-/css-color-3.2.0.tgz#cc42f5b85c593f79f1fa4f25d2b9b321e61d1794"
```
adumesny, line 7: guess you're using npm instead of yarn, which I guess is fine...

adumesny (approved):
thank you for the changes. but those will get blown away when the p[ending React changelist comes in, so wondering if you could look at https://github.com/gridstack/gridstack.js/pull/2938 as well ?

PR #3159, opened 2025-09-16
Fix bug where removeWidget does not function properly

YOUNGHO8762:
### Description
resolving the issue where only the internal elements were removed when an ID was provided. Passing the element now ensures that the entire grid-stack-item-content is removed as intended.

### Checklist


adumesny:
thanks for the fix. alas the compiler doesn't catch that issue since passing a string is valid, btu looks for css selectors, not gs-id attributes (which it could I suppose as fallback).

PR #3163, opened 2025-09-22
Esc doesn't restore subgrid (partial)

### LINKED ISSUE #3154: [Bug] Cancelling when dragging into a subgrid removes the widget instead of restoring the origin state
## Subject of the issue
When dragging a widget from 1 grid into another and you press the cancel key (`escape`); the widget just dissapears. This is also the case in the nested grid demo: https://gridstackjs.com/demo/nested.html#

## Your environment
* Currently running gridstack: 12.2.2 in an Angular Project

## Steps to reproduce
https://gridstackjs.com/demo/nested.html#
- Grab widget with the number 2 on it and drag it to another subgrid
- Press the `escape` key

## Expected behavior
When pressing the `escape` key the widget should return to its original location instead of disapearing

adumesny:
### Description
* partial fix #3154
* started looking at fixing nested changes to unroll but it turn out to be a lot more complex - saving those WIP for another time...

### Checklist

### DIFF
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -331,7 +331,7 @@ export class DDDraggable extends DDBaseImplement implements HTMLElementExtendOpt
   /** @internal restore back the original style before dragging */
   protected _removeHelperStyle(): DDDraggable {
     this.helper.classList.remove('ui-draggable-dragging');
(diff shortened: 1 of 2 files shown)


adumesny:
have another fix

PR #3165, opened 2025-09-23
react wrapper grid stack provider API to accept widget objects directly

YOUNGHO8762:
### Description
This PR updates the GridStackProvider API to accept widget objects directly instead of callback functions, simplifying the interface for adding widgets and subgrids.

**Changes**

- Modified `addWidget` implementation to accept widget object directly and use provided ID
- Modified `addSubGrid` implementation to accept subgrid object with children array
- Removed internal ID generation logic and callback-based widget creation

**Benefits**

- Simplified API usage - no need for callback functions
- Better control over widget IDs for consumers
- More predictable widget creation process
- Cleaner type definitions and implementation

### Checklist

### DIFF
--- a/react/lib/grid-stack-context.ts
+++ b/react/lib/grid-stack-context.ts
@@ -4,13 +4,15 @@ import { createContext, useContext } from "react";
 export const GridStackContext = createContext<{
   initialOptions: GridStackOptions;
   gridStack: GridStack | null;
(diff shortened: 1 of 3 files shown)


adumesny:
I like that it uses the GS methods directly with little extra code. If you look at the Angualr wrapper, you can still access all API directly from GridComponent.grid and you get callback to create the angular wrappers needed, btu GS API can be used directly.... no need to dup the API and doc is the same.

PR #3169, opened 2025-10-02
fix: only show resize handles on hover in Chromium browsers when auto…

kishansinghifs1:
# This PR adds a CSS fallback to ensure that resize handles are only visible on hover when autohide is enabled, fixing an issue where Chromium-based browsers (Chrome, Edge) always showed the handles regardless of hover state.


adumesny:
closing due to invalide code

PR #3191, opened 2025-11-17
Fixed touch issue on Linux

jemunk:
### Description
Fixed this touch issue happening on Linux device with touch screen running the application under Electron (Chromium):
When user starts dragging an item, as soon finger leaves the element, _mouseLeave is run, called from @dd-touch.js simulating mouse event. But on next move, _mouseEnter is run even though pointer has not reentered the element. _mouseEnter is not called by @dd-touch.js as part of simulating mouse, but is called as a real mouse event by browser. Strange part is that event.screenX from the _mouseEvent argument, gives a coordinate that matches where finger started dragging, not the coordinate of where the finger left the element.

### Checklist


adumesny:
thanks for the fix!

jemunk:
Thanks for merging. Do you have an idea of when it will be available on NPM?

adumesny:
@jemunk please verify the correct fix (only tested in simulation) which will go out in 12.4.2

Issue #819, opened 2017-11-22
gridstack rtl mode


BrkCoder:
## Subject of the issue
When moving to rtl mode, the layout doesn't organize in rtl direction rather in ltr direction although I have set the flag of rtl options to true.

## Your environment
version of gridstack is 1.0.0

## Steps to reproduce
in-order to reproduce the problem all we need is the configure the following options for gridstack:
options = {
            cellHeight: 100,
            verticalMargin: 10,
            draggable: {
                handle: '.header-drag',
            },
            rtl: true
}

## Expected behaviour
Expected behavior widgets start adding from right to left. 

## Actual behaviour
Actual behavior widgets start adding from left to right.

MrMowji:
Unfortunately that's not the only behavior. Re-sizable handlers should be on the left side of `grid-stack-item`.

radiolips:
Sounds like #820 might be this same issue? This might be tricky, but it's a reasonable request.

sinap-se:
Rough PoC:

if (this.opts.rtl) {
        for (var i = 0;; ++i) {
          var x = this.column - (i % this.column) - node.width;
          var y = Math.floor(i / this.column);
          if (x + node.width > this.column || x < 0) {
            continue;
          }
          if (!this.nodes.find(Utils._isAddNodeIntercepted, {x: x, y: y, node: node})) {
            node.x = x;
            node.y = y;
            delete node.autoPosition; // found our slot
            break;
          }
        }
      } else {
        for (var i = 0;; ++i) {
          var x = i % this.column;
          var y = Math.floor(i / this.column);
          if (x + node.width > this.column) {
            continue;
          }
          if (!this.nodes.find(Utils._isAddNodeIntercepted, {x: x, y: y, node: node})) {
            node.x = x;
            node.y = y;
            delete node.autoPosition; // found our slot
            break;
          }
        }

adumesny:
@sinap-se do you mind making a review fix in the latest 3.0 code for this ? (typescript now)
I would love to get this in an official build. We can update the rtl demo as well as I have not spent time on this.

I can take care of the default resize handle side.

adumesny:
actually wondering if this could all be done in CSS instead as column 0 would be right most in rtl mode,  and all position is done in CSS, rather than change code... but maybe not as width would need to be subtracted from position. Would be interesting to see what browser support it (ideally coordinate and width would be flipped and not require code change).

Vishnupriyaakinapelli:
[...]

PR #3250, opened 2026-03-30
implemented full rtl support

danielgindi:
### Description
This PR completes the RTL behavior in such a way that:
1. Elements flow from right to left when `rtl` is set. The original `Position` struct has not changed, it still has a `left` property, which now means "horizontal start position". When in rtl, it will be the position from the right.
2. Dragging - works correctly in rtl mode. With minimal changes to the code.
3. Resizing - the same. The resize event has new properties to make it more semantic without testing for the `rtl` flag everywhere.

This closes the issue #819.

### Checklist
- [ ] Created tests which fail without the change (if possible) - *not possible, tests already fail on `master`*
- [ ] All tests passing (`yarn test`) - *the same tests fail on `master`*
- [ ] Extended the README / documentation, if necessary - *not necessary*

### DIFF
--- a/src/dd-draggable.ts
+++ b/src/dd-draggable.ts
@@ -12,11 +12,11 @@ import { isTouch, touchend, touchmove, touchstart, pointerdown, DDTouch } from '
 import { GridHTMLElement } from './gridstack';
 
 interface DragOffset {
-  left: number;
+  x: number;
   top: number;
   width: number;
   height: number;
-  offsetLeft: number;
+  offsetX: number;
   offsetTop: number;
 }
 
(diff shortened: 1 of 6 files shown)


adumesny (commented):
thank you for making the fixes! I have a few small tweaks

danielgindi:
> thank you for making the fixes! I have a few small tweaks

I resolved those comments

adumesny (approved):
thank you for the feature!

danielgindi:
Btw, @adumesny, I believe that we should get rid of `direction: rtl` in the css, as the user may want an rtl direction for the grid, without affecting the css of the item, and vice versa. Let the user control the css with css :-)

adumesny:
@danielgindi sure that sounds fine. just update the demo and any readme then. I'm not that familiar nor use rtl

PR #3267, opened 2026-04-11
updateOptions() fixes for maxRow when set row

evovas:
### Description
This is the same change that was made in [#3237](https://github.com/gridstack/gridstack.js/pull/3237) for maxRow, but for the case where we specify an exact number of rows.

### Checklist

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -1508,7 +1508,7 @@ export class GridStack {
     if (o.disableResize !== undefined && !o.staticGrid) this.enableResize(!o.disableResize);
     if (o.float !== undefined) this.float(o.float);
     if (o.row !== undefined) {
-      opts.minRow = opts.maxRow = opts.row = o.row;
+      opts.minRow = opts.maxRow = this.engine.maxRow = opts.row = o.row;
       this._updateContainerHeight();
     } else {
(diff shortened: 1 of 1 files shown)


src/gridstack.ts
adumesny, line 1515: you also need it here

evovas: Coincidentally, I encountered a bug where the row count wasn't updating when changing options, right when an update was released that fixed this issue.

But in my code, I wasn't updating `o.maxRow` and `o.minRow`, but specifically one value, `o.row`. Simply updating the library to the latest version didn't help, but after examining the source code, I realized I could replace the code block from this:
```
grid.updateOptions({
  ...
  row: 4,
  ...
})
```
to this 
```
grid.updateOptions({
  ...
  minRow: 4,
  maxRow: 4,
  ...
})
```
And this changes fixed the issue for me.

But I think that this change can be supported even when specifying a specific number of rows via `o.row`, but now the code is executed in such a way that if `o.row` is specified, then the block in `else`, in which you made the behavior corrections, is not executed.

adumesny: nevermind, I didn't see I had made the change. yes thanks for the fix.

adumesny (approved):

PR #3270, opened 2026-04-15
move resizable handles' styling to css, higher z-index for corners

### LINKED ISSUE #819: gridstack rtl mode

danielgindi:
### Description
The handles had their css set in HTML, which was both dirty and forcing the use of `!important` rules in css in order to modify.
I moved those to css, and set better defaults for z-index (corners should be higher than sides, as they have less grabbing space and should not be obscured by the side handles).

* fix #819

### Checklist

### DIFF
--- a/src/dd-resizable-handle.ts
+++ b/src/dd-resizable-handle.ts
@@ -53,8 +53,6 @@ export class DDResizableHandle {
 
     this.el.classList.add('ui-resizable-handle');
     this.el.classList.add(`${DDResizableHandle.prefix}${this.dir}`);
-    this.el.style.zIndex = '100';
-    this.el.style.userSelect = 'none';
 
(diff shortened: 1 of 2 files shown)


adumesny (approved):
thanks for cleanup!

PR #3360, opened 2026-08-18
Fix resizing with north and west resize handles

ryanpetebyte:
### Description
I changed the _dragOrResize method to change the x and y positions to be calculated by the width and height change, specifically only for the north and west handles. For example the original method could cause the widgets to grow in width, but keep the same x depending on the resolution and the position of the cursor.

### Checklist
- [ ] All tests passing (`yarn test`) (I don't know why, but there were several failures when I ran `yarn lint`, and none in what I changed. I've never done .ts development so my system might not be configured correctly.)

### DIFF
--- a/src/gridstack.ts
+++ b/src/gridstack.ts
@@ -3084,7 +3084,7 @@ export class GridStack {
       // for SE/S/E handles the top-left is anchored — recalculating from pixels causes
       // rounding drift on fine grids where cellWidth/cellHeight are only a few pixels. #385 #1356
       if (event.hasMovedX) {
-        // Set the X based on the node properties.
+        // Set the X based on already calculated p width and node properties.
         if (node.x != undefined && p.w != undefined && node.w != undefined) {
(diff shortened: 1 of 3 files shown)


src/gridstack.ts
```diff
@@ -3084,12 +3084,26 @@ export class GridStack {
       // for SE/S/E handles the top-left is anchored — recalculating from pixels causes
       // rounding drift on fine grids where cellWidth/cellHeight are only a few pixels. #385 #1356
       if (event.hasMovedX) {
-        const left = ui.position!.left + mLeft;
-        p.x = Math.round(left / cellWidth);
+        // Set the X based on already calculated p width and node properties.
+        if (node.x != undefined && p.w != undefined && node.w != undefined) {
```
adumesny, line 3088: would those ever be undefined ? seem like using the w/h we just calculated above is simpler and right way to do this, but I have to test why old code wasn't working in your case...

